PR #5, opened 2015-10-06
Sprint 32 TLT-1312 Add server side processing for paging/sorting

douglashall:
* This PR depends on https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/11
* Adds server-side paging and ordering of the course instance datatable.


elliottyates:
@elliott

course_info/static/course_info/js/controllers/IndexController.js
```diff
@@ -124,9 +142,55 @@
...
+                        success: function(data, textStatus, jqXHR) {
+                            var results = data.results;
+                            var resultsLength = data.results.length;
+                            var processedData = [];
+                            for (var i = 0; i < resultsLength; i++) {
+                                processedData.push($scope.courseInstanceToTable(results[i]));
+                            }
+                            callback({
+                                recordsTotal: data.count,
+                                recordsFiltered: data.count,
```
elliottyates: I guess this means that recordsFiltered/recordsTotal distinction doesn't make much sense in our context, since we're not filtering within a larger dataset, but we have to return this as part of datatables I suppose? A comment in here about how our API treats these might be nice, since a useful recordsTotal won't be returned by the API it seems.

course_info/static/course_info/js/controllers/IndexController.js
```diff
@@ -75,6 +90,9 @@
 
         $scope.checkIfSearchable = function() {
             $scope.searchEnabled = $scope.filtersApplied || $scope.queryString.trim() != '';
+            if (!$scope.searchEnabled) {
```
elliottyates: This used to be predicated upon `searchInProgress`, not `searchEnabled`. So this means that if someone removes filters and search values, the data table disappears, even if they haven't initiated a new search?

elliottyates:
:+1: looks good!

sapnamysore:
Started looking at this PR earlier - giving a cursory glance at the rest  of the stuff, as I see @elliottyates  has picked up the PR. Looks good to me

course_info/static/course_info/js/controllers/IndexController.js
sapnamysore: hmm, can a useful  recordTotal be derived from the 'count'? [https://icommons.harvard.edu/api/course/v2/course_instances/]

elliottyates: I think our API returns `count` as the number of records found after applying the query parameters. DataTables has an ability to show records filtered (I don't think we're enabling it in this UI), but it's ambiguous whether when using serverside processing that returned `count` should be recordsFiltered or recordsTotal (which might be, say, the total count of all CourseInstances unfiltered by the query parameters). I think recordsFiltered makes more sense when doing client-side filtering (because it should be effectively instantaneous), but really it's the same thing in the end, the filtering is just being applied in different places.

douglashall: For our use of DataTables here recordsFiltered = recordsTotal = count (which is what is returned from the API call). As Elliott laid out, you can have DataTables display something like "Showing 1 to 2 of 2 entries (filtered from 57 total entries)" by setting recordsFiltered and recordsTotal to different values. DataTables does this automatically when you are not using server-side processing and you use the out of the box search input that DataTables provides. Again for our use case recordsFiltered should always be equal to recordsTotal.

course_info/static/course_info/js/controllers/IndexController.js
douglashall: Yes, that is correct. These 3 lines can be removed if that is not desired.

PR #6, opened 2015-10-07
Sprint 32 TLT-1312 Clean up icommons_rest_api proxy

douglashall:
* Strip resource_link_id query param from requests made to icommons_rest_api
* Ensure a trailing slash on requests made to icommons_rest_api


elliottyates:
:+1:

PR #9, opened 2015-10-08
Sprint 33 TLT-1312 Require icommons_rest_api package and use it for course info search

douglashall:
Depends on PR:

https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/14

### DIFF
--- a/canvas_account_admin_tools/urls.py
+++ b/canvas_account_admin_tools/urls.py
@@ -10,7 +10,7 @@ urlpatterns = [
     url(r'^not_authorized$', icommons_ui_views.not_authorized, name='not_authorized'),
     url(r'^tool_config$', views.tool_config, name='tool_config'),
     url(r'^lti_launch$', views.lti_launch, name='lti_launch'),
-    url(r'^icommons_rest_api/(?P<path>.*)$', views.icommons_rest_api_proxy, name='icommons_rest_api_proxy'),
     url(r'^account_dashboard$', views.dashboard_account, name='dashboard_account'),
     url(r'^course_info/', include('course_info.urls', namespace='course_info')),
+    url(r'^api/course/v2/', include('course_api.urls')),
 ]
--- a/course_info/templates/course_info/index.html
+++ b/course_info/templates/course_info/index.html
@@ -16,6 +16,7 @@
     <script src="{% static 'datatables-bootstrap3/BS3/assets/js/datatables.js' %}"></script>
     <script>
         window.globals.STATIC_URL = '{% settings_value "STATIC_URL" %}';
+        window.globals.COURSE_API_URL = '{% url "courseinstance-list" %}';
     </script>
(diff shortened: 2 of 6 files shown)


canvas_account_admin_tools/urls.py
```diff
@@ -10,7 +10,7 @@
     url(r'^not_authorized$', icommons_ui_views.not_authorized, name='not_authorized'),
     url(r'^tool_config$', views.tool_config, name='tool_config'),
     url(r'^lti_launch$', views.lti_launch, name='lti_launch'),
-    url(r'^icommons_rest_api/(?P<path>.*)$', views.icommons_rest_api_proxy, name='icommons_rest_api_proxy'),
     url(r'^account_dashboard$', views.dashboard_account, name='dashboard_account'),
     url(r'^course_info/', include('course_info.urls', namespace='course_info')),
+    url(r'^api/course/v2/', include('course_api.urls')),
```
rascalking: Any reason not to keep this at `/icommons_rest_api`, rather than a generic `/api`?

rascalking: Also, can we namespace this like the other apps?

douglashall: Yeah, that was causing problems when DRF tries to generate the resource URLs that it includes in responses.

course_info/templates/course_info/index.html
```diff
@@ -16,6 +16,7 @@
     <script src="{% static 'datatables-bootstrap3/BS3/assets/js/datatables.js' %}"></script>
     <script>
         window.globals.STATIC_URL = '{% settings_value "STATIC_URL" %}';
+        window.globals.COURSE_API_URL = '{% url "courseinstance-list" %}';
```
rascalking: Sorry, I just landed a change that pulls the term code dropdown data from the api as well...we need to update this to point to the api root, then append the resource name in the ajax call.

douglashall: Got it.

canvas_account_admin_tools/urls.py
rascalking: Fantastic.  And it looks like that's not something we can fix, unless we want to update icommons_rest_api and mandate that it's always included under a set namespace.

http://www.django-rest-framework.org/api-guide/routers/#usage
(sadly no link directly to it, just search for "namespace" in page)

elliottyates: Any reason why we wouldn't update the icommons_rest_api to a more generic namespace as part of this change, since we're moving towards it as a more robust replacement for the existing one anyway? I guess it would be hard to figure out what namespace works, in a case where it would be included with other APIs in a project, but if we're moving towards running it in its own project context eventually then that's not a big issue...?

elliottyates:
See Oct 13 #devops channel slack chat. Closing for now.

PR #12, opened 2015-10-09
Sprint 33 Use sqlite for tests, add db app mapping

rascalking:
All in an attempt to get unit tests working under jenkins.


douglashall:
:+1:

PR #13, opened 2015-10-15
Task/elliottyates/tlt 2025/course info ui fixes

elliottyates:
- restricts access to admin dashboard to LTI administrator roles only
- default sorting is by the Course Details column in ascending order
- For the BAs, the default selection is "All schools" and the current year
- For the liaisons, the default selection is their school (and associated schools, if any), and the current year
- The search button is always active
- Non cross listed courses don't have any Cross Listing label, rather an "N/A"
- Courses without sites show an "N/A" in place of the site link
- column header capitalization
- disables column sorting (i.e. a new search) while an existing search is underway on course_info index page

### DIFF
--- a/canvas_account_admin_tools/views.py
+++ b/canvas_account_admin_tools/views.py
@@ -14,7 +14,8 @@ from ims_lti_py.tool_config import ToolConfig
 from canvas_account_admin_tools.models import ExternalTool
 
 from proxy.views import proxy_view
-
+from django_auth_lti import const
+from django_auth_lti.decorators import lti_role_required
 
 logger = logging.getLogger(__name__)
 
@@ -54,6 +55,8 @@ def lti_launch(request):
 
 
 @login_required
+@lti_role_required(const.ADMINISTRATOR)
+@require_http_methods(['GET'])
 def dashboard_account(request):
     custom_canvas_account_id = request.LTI['custom_canvas_account_id']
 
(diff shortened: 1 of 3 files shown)


sapnamysore:
For this  criteria: "For the liaisons, the default selection is their school (and associated schools, if any), and the current year" 
- is this just the default behavior bcoz the liaisons would have only one school? I don't see any changes explicitly for this but do see that the first item from the schools list is selected.

sapnamysore:
Also, the it looks like the value 'All Schools' is inserted to the schools list for everyone. Would that work correctly for the liaisons?

sapnamysore:
Other than the clarification about school list, the changes look good :+1:

elliottyates:
Ah you're right about All Schools still being default on liaisons, good catch. I thought that when the school list was filtered the first one on the list would be the liaison's school. Need to think about how to make that happen then.

Vittorio2015:
I believe that it is also possible that a liaison may have access to more than one school. In any case, their primary school shall be the default selection.

Vittorio

elliottyates:
The 'all schools' option issue will be pulled out into https://jira.huit.harvard.edu/browse/TLT-2093 and https://jira.huit.harvard.edu/browse/TLT-2094, and https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/15 removes the option entirely.

PR #14, opened 2015-10-19
Task/sapnamysore/tlt 2056/selenium tests

sapnamysore:
This PR contains selenium tests for course info app. Most of the files are within the selenium_tests folder with some configurations in the settings. The navigation to the course info page is via the canvas_account_admin_tools landing page-  the course_info_parent_page_object refers to  this landing page. 

@elliottyates , @Joy603 , @hktest


elliottyates:
@elliottyates

canvas_account_admin_tools/settings/local.py
```diff
@@ -18,4 +18,14 @@
...
 # with an unverified certificate, e.g. the one that ships with django-sslserver
-ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = True
\ No newline at end of file
+ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = True
+
+SELENIUM_CONFIG = {
+   'selenium_username': SECURE_SETTINGS.get('selenium_user'),
+   'selenium_password': SECURE_SETTINGS.get('selenium_password'),
+   'selenium_grid_url': SECURE_SETTINGS.get('selenium_grid_url'),
+   'base_url': 'https://canvas.icommons.harvard.edu',
```
elliottyates: Do we need `base_url`? Isn't the `canvas_base_url` achieving the same thing?

elliottyates: (don't see it used, so it's probably safe to remove it)

selenium_tests/base_selenium_test_case.py
```diff
@@ -0,0 +1,48 @@
...
+    driver = None  # make selenium driver available to any part of the test case
+    display = None  # a reference to the virtual display (for running tests locally)
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Sets up the test case, including the selenium browser driver to use
+        """
+
+        local = True
```
elliottyates: This should use the SELENIUM_CONFIG variable now available. Some of the comments in this file can be removed too (see the most recent updates to this file in Hiu Kei's PR), but that's not important.

selenium_tests/course_info/course_info_base_test_case.py
```diff
@@ -0,0 +1,37 @@
...
+        cls.course_info_parent_page.get(cls.TOOL_URL)
+        login_page = PinLoginPageObject(driver)
+        if login_page.is_loaded():
+            print " logging the user in"
+            login_page.login(cls.USERNAME, cls.PASSWORD)
+        else:
+            print '(User {} already logged in to PIN)'.format(cls.USERNAME)
+
+    @classmethod
+    def tearDownClass(cls):
```
elliottyates: If this isn't implemented, won't the `super` automatically get called? I guess if we're not adding any functionality here we can remove this, right? (It may be in some of our files as a placeholder / reminder, but we don't seem to need it right now.)

selenium_tests/regression_suite.py
```diff
@@ -0,0 +1,52 @@
...
+> python selenium_tests/canvas_account_admin_tools/test_search.py
+
+In PyCharm, if xvfb is installed already, you can run them through the Python unit test run config
+(make sure the above environment settings are included)
+"""
+
+import unittest
+import time
+
+# from selenium_tests.bulk_create.bc_test_flow import TestFlow
```
elliottyates: Looks like this can be removed

selenium_tests/pin/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'sam576'
```
elliottyates: Can remove these (it's a PyCharm default, can be changed via Preferences > IDE Settings > File and Code Templates).

selenium_tests/pin/page_objects/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'sam576'
```
elliottyates: ditto

selenium_tests/course_info/course_info_is_search_page_loaded_test.py
```diff
@@ -0,0 +1,28 @@
...
+from selenium_tests.course_info.page_objects.course_info_search_page_object import CourseSearchPageObject
+
+
+class CourseInfoIsSearchPageLoadedTest(CourseInfoBaseTestCase):
+
+    def test_is_loaded(self):
+        """
+        Verify  that the course info search page is loaded by checking against page title
+        """
+        driver = self.driver
```
elliottyates: I would prefer to see all of the `driver = self.driver` lines removed (in the other tests too), and `self.driver` used in place of `driver`. Tightens things up a bit.

selenium_tests/course_info/course_info_is_search_page_loaded_test.py
```diff
@@ -0,0 +1,28 @@
...
+        parent_page = CourseInfoParentPage(driver)
+        search_page = CourseSearchPageObject(driver)
+
+        # navigate to course info search page
+        search_page = parent_page.select_course_info_link()
+        element = search_page.get_page_title()
```
elliottyates: You can simply `self.assertTrue(search_page.is_loaded())` I think; the test shouldn't need to know the details of the page's title, etc -- the page object is the expert there.

selenium_tests/course_info/page_objects/course_info_parent_page_object.py
```diff
@@ -0,0 +1,52 @@
...
+
+        if title and 'Administration Tasks' in title.get_attribute('textContent'):
+            return True
+        else:
+            raise RuntimeError(
+                'Could not determine if Course Info  base page loaded as expected;'
+                'title element was found but did not contain expected text'
+            )
+
+    def get_page_title(self):
```
elliottyates: I think this can be removed entirely.

selenium_tests/course_info/course_info_is_initialized_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def test_is_link_present(self):
+        """
+        Check that Course Info link is loaded by checking against Link text
+        """
+        driver = self.driver
+        page = CourseInfoParentPage(driver)
+
+        # verify that the course info block is present
+        course_info_link = page.get_course_info_link()
```
elliottyates: I think it would be better to encapsulate this in the page object, e.g. have the page object present a method which checks this stuff (page element details, etc), and here all that's needed is `self.assertTrue(is_course_info_block_present())`.

selenium_tests/course_info/course_info_is_initialized_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def test_is_page_loaded(self):
+
+        """
+        Check if page is loaded
+        """
+        driver = self.driver
+        # initialize
+        page = CourseInfoParentPage(driver)
+        page.is_loaded()
```
elliottyates: No assert in this test?.... Also the test name is a bit confusing -- this appears to check whether the parent page is loaded, but that isn't clear to me from the class name + method name combination.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -0,0 +1,72 @@
...
+
+class CourseSearchPageObject(CourseInfoBasePageObject):
+
+    def is_loaded(self):
+        """ determine if the page loaded by looking for a specific element on the page """
+        # frame context stickiness is a bit flaky for some reason; make sure we're in the tool_content frame context
+        # before checking for elements on the expected
+        self.focus_on_tool_frame()
+        try:
+            self.find_element(*CourseSearchPageLocators.COURSE_INFO_SEARCH_BUTTON)
```
elliottyates: This doesn't seem like a very selective selector -- I would imagine other pages where the driver would find this object, but it might not be the Course Search Page. It appears to be only looking for a button.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -0,0 +1,72 @@
...
+        except NoSuchElementException:
+            return False
+        return True
+
+    def get_page_title(self):
```
elliottyates: This can be removed (and `is_loaded()` used instead, in the test case).

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -0,0 +1,72 @@
...
+        :returns boolean
+        """
+        search_button = self.find_element(*CourseSearchPageLocators.COURSE_INFO_SEARCH_BUTTON_PATH)
+        return search_button.is_enabled()
+
+    def submit_search(self, search_text):
+        self.focus_on_tool_frame()
+        self.find_element(*CourseSearchPageLocators.SELECT_SCHOOL_DROPDOWN).click()
+        driver = self._driver
+        driver.find_element_by_xpath("//input[@type='text']").clear()
```
elliottyates: Can these be factored into `LOCATOR`s and described by their locator variable name, so it's clearer what's happening?

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -0,0 +1,72 @@
...
+            return False
+        return True
+
+    def get_page_title(self):
+        element = self.find_element(*CourseSearchPageLocators.PAGE_TITLE)
+        return element
+
+    def select_school(self, school):
+        """ select a school from the schools dropdown """
+        self.focus_on_tool_frame()
```
elliottyates: It seems like the test should do an `is_loaded()` check on the PO if you want to focus on the tool frame... otherwise we have to add this into every `select_*()` method here (what if the test decided to load the page then use `select_year()` first, instead of `select_school()`)?

selenium_tests/course_info/course_info_search_test.py
```diff
@@ -0,0 +1,45 @@
...
+        search_page.submit_search("339331")
+
+        try:
+            WebDriverWait(driver, 30).until(lambda s: s.find_element(
+                *CourseSearchPageLocators.COURSE_RESULTS_TABLE).is_displayed())
+        except TimeoutException:
+            return False
+
+        body_text = self.driver.find_element_by_tag_name('body').text
+        print(body_text)
```
elliottyates: Probably not needed anymore

selenium_tests/course_info/course_info_is_search_page_loaded_test.py
```diff
@@ -0,0 +1,28 @@
...
+
+class CourseInfoIsSearchPageLoadedTest(CourseInfoBaseTestCase):
+
+    def test_is_loaded(self):
+        """
+        Verify  that the course info search page is loaded by checking against page title
+        """
+        driver = self.driver
+        parent_page = CourseInfoParentPage(driver)
+        search_page = CourseSearchPageObject(driver)
```
elliottyates: The `select_course_info_link()` call below (line 19) is returning the PO for you, so I think this line isn't needed.

elliottyates:
I would consider pulling the course_info_parent_page_object.py out into its own e.g. admin_dashboard_page_object.py, as that's going to provide services for all the other apps that we'll need to test in this project. So the course_info tests will load that page object as a starting point, click on the course_info link, which will return a course_info page object, etc.

canvas_account_admin_tools/settings/local.py
sapnamysore: Good catch. thanks

selenium_tests/account_admin/account_admin_base_test_case.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        setup values for the tests
+        """
+        super(AccountAdminBaseTestCase, cls).setUpClass()
+
+        driver = cls.driver
+        cls.USERNAME = settings.SELENIUM_CONFIG.get('selenium_username')
+        cls.PASSWORD = settings.SELENIUM_CONFIG.get('selenium_password')
+        cls.CANVAS_BASE_URL = settings.SELENIUM_CONFIG.get('canvas_base_url')
+        cls.TOOL_RELATIVE_URL = settings.SELENIUM_CONFIG.get('course_info_tool_relative_url')
```
elliottyates: This should be its own selenium_config setting, given the test suite for this page is independent of the test suite for the course_info tool.

elliottyates: (the setting will wind up having the same value, for now, but there is a cleaner separation that way)

elliottyates:
Ok just the comment about the new `AccountAdminBaseTestCase` `TOOL_RELATIVE_URL`, and there was one comment still hanging from the first PR: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/14#discussion_r42388554

elliottyates:
Otherwise, :+1:

selenium_tests/account_admin/account_admin_base_test_case.py
sapnamysore: Thanks. Yes, makes sense

PR #15, opened 2015-10-19
Sprint 33 TLT-1312 Begin adding Jasmine js unit tests

rascalking:
* Removes "All Schools" filter option.
* Sets up the skeleton of karma-driven jasmine unit tests for the index controller.
* Passes the list of schools via `window.schoolOptions` set in the head, rather than a json blob in the body.
* Removes unneeded `updateFilter()` method.
* Switches to using `$window` and `$document` for better testability.


elliottyates:
:+1:

PR #16, opened 2015-10-19
Sprint 33 TLT-1312 Handle new term_codes return structure

rascalking:
Missed this on the jasmine branch, sorry.

PR #17, opened 2015-10-20
Revert "Sprint 33 TLT-1312 Handle new term_codes return structure"

rascalking:
Looks like we had an old version of the rest api running on dev, and this change should never have been made.

Reverts Harvard-University-iCommons/canvas_account_admin_tools#16


sapnamysore:
:+1: 

Does this break anything in the Jasmine tests?

rascalking:
Nope.

PR #20, opened 2015-11-14
Fix jasmine tests

rascalking:
* there's no such thing as scope.new(), it's .$new().
* update the term_codes url and response to match what we're using now
* add phantomjs as dev dependency


elliottyates:
:+1:

PR #24, opened 2015-12-01
Sprint 36 TLT-2121 Fix manual/registrar logic

rascalking:
Fix the regex so that our "was it manually or registrar added" logic matches what's in manage people.


elliottyates:
:+1:

PR #26, opened 2015-12-03
Sprint 37 TLT-2121 Add people view jasmine tests

rascalking:
The interesting bits here are the directive/service/controller tests.

Note: Corrects the route/controller/template names from enrollments to people, since the view includes teachers/tas/etc.


elliottyates:
@elliott

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -85,33 +100,27 @@
                 },
             },
             lengthChange: false,
-            sAjaxDataProp: 'aaData',
+            sAjaxDataProp: 'data',
```
elliottyates: This is listed as a legacy option (like all the type-prefixed stuff in datatables); do we still need it despite having `dataSrc` defined in the `ajax` callback? If so, maybe another comment would be helpful.

course_info/tests/jasmine/directives/BadgeDirectiveSpec.js
```diff
@@ -0,0 +1,45 @@
...
+    // sanity check the test framework
+    it('should inject the providers we requested', function() {
+        [$compile, $rootScope].forEach(function(thing) {
+            expect(thing).not.toBeUndefined();
+            expect(thing).not.toBeNull();
+        });
+    });
+
+    // these are just copied from BadgeDirective.js, and should be kept in sync
+    var roleToBadge = {
```
elliottyates: We can't share these somehow across modules? Too much trouble, perhaps?

course_info/tests/jasmine/directives/BadgeDirectiveSpec.js
```diff
@@ -0,0 +1,45 @@
...
+        XIDHOLDER: 'XID',
+    };
+    var badgeToClass = {
+        HUID: 'label-danger',
+        XID: 'label-primary',
+        LIBRARY: 'label-success',
+        OTHER: 'label-warning',
+    };
+
+    for (role in roleToBadge) {
```
elliottyates: Perhaps one more test for when the role isn't in `roleToBadge` (i.e. test for `OTHER`)

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -0,0 +1,125 @@
...
+            $routeParams = _$routeParams_;
+            courseInstances = _courseInstances_;
+            $compile = _$compile_;
+            djangoUrl = _djangoUrl_;
+            $httpBackend = _$httpBackend_;
+            $window = _$window_;
+
+            // this comes from django_auth_lti, just stub it out so that the $httpBackend
+            // sanity checks in afterEach() don't fail
+            $window.globals = {
```
elliottyates: It kinda sucks that we have to do this everywhere there's an ajax request, huh.... I see we're not going to do the `verifyNoOutstandingExpectation()` etc checks everywhere?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -0,0 +1,125 @@
...
+    describe('setTitle', function() {
+        var ci;
+        beforeEach(function() {
+            ci = {
+                course_instance_id: $routeParams.course_instance_id,
+                title: 'Test Title',
+            };
+            courseInstances.instances = {};
+        });
+        afterEach(function() { courseInstances.instances = {}; });
```
elliottyates: Why do we have to clear the object before and after each `it()`?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -0,0 +1,125 @@
...
+            expect(result).toBe('Student, Joe');
+        });
+
+        it('renderId', function() {
+            var full = {
+                profile: {role_type_cd: 'STUDENT'},
+                user_id: 123456,
+            }
+            var result = scope.renderId(undefined, undefined, full, undefined);
+            expect(result).toBe('<badge ng-cloak role="STUDENT"></badge> 123456');
```
elliottyates: Pretty bizarre. So angular interprets this after it's shown? I'm beginning to dislike Angular. :)

elliottyates:
:+1:

course_info/static/course_info/js/controllers/PeopleController.js
rascalking: Sadly, yes.  If you use the function version of the `ajax` argument, and you're using server-side processing (since that requires returning an object with the rowsFiltered/rowsReturned attrs, rather than just an array of data), your only option to make sure the data gets displayed is using the deprecated `sAjaxDataProp` param.

I found a datatables forum post where one of the devs confirmed it was a bug at one point, I'll see if I can scare it up for a comment.

course_info/tests/jasmine/directives/BadgeDirectiveSpec.js
rascalking: Nah, I just err on the side of not sharing code between tests and the code under test if I can help it.  Happy to do it that way if you'd prefer.

course_info/tests/jasmine/directives/BadgeDirectiveSpec.js
rascalking: Good idea.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yeah...it's worth investigating a base class for ajax tests the next time around.  And, hell...forgot about the `verify` calls...adding them in.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: We might not...I was being paranoid about resetting the injected service, but now that I think about it, I think jasmine is injecting a new copy of it for every test...I'll confirm, and hopefully yank these.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: This is just testing that our render function is returning the correct string, right?  At no point have we tried to hang this off the dom.

But yes, for dom elements dynamically generated by other sources (ie. datatables), you have to manually [compile them](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/26/files#diff-e045fa70e4df999f02b9ae1ebead90b8R91) yourself.

The really awesome part about that is that the compiled dom (ie. what angular replaces your custom tag with) isn't simply replaced...it's inserted as child elements of your custom tag.  That's why I'm searching for children [here](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/26/files#diff-5f5c4f29557cabab2a819964f13191b3R41), instead of just checking the tag name.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yeah, don't need this.

PR #27, opened 2015-12-14
filter user list by active status and prime_role_indicator

eparker71:
filter user list by active status and prime_role_indicator

I make a dict of the id's and put the records asssociated with the id in a list.
then I sort the list for each id and grab the top one. 

I created a custom comparator function to handle the sorting.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -67,8 +67,55 @@
                     else if (data.results.length === 1) {
                         // TODO - POST user, handle results
                     }
+
+                    $scope.searchResults = $scope.filterResults(data.results);
                 })
                 .error($scope.ajaxErrorHandler);
+        };
+        /*
+            concat the active flag with the prime_role_indicator
+            value to be able to sort records base on these values
+            active = (0 | 1) a value of 1 here trumps the prime_role_indicator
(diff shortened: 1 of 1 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -67,9 +67,56 @@
...
+            active = (0 | 1) a value of 1 here trumps the prime_role_indicator
+            prime_role_indicator = ( "Y" | "N" | "")
+            1 > 0  true
+            '1:string' > '0:string' true
+            'Y' > 'N' true
+            This should let any records with a 1 in the active column float to the top
+            if there are non, records with a Y in the prime_role_indicator will float up
+            and records with both will float above each of those.
+         */
+        $scope.compare = function(a, b) {
```
elliottyates: A more concise way, that doesn't involve concatenation, would be something like:
```
$scope.compare = function(a, b) {
  return b.active == a.active 
    ? b.prime_role_indicator > a.prime_role_indicator 
    : b.active > a.active;
};
```

elliottyates:
:+1:

course_info/static/course_info/js/controllers/PeopleController.js
eparker71: I'm usually not in favor of things that are visually less readable, but this is nice!! :)

PR #28, opened 2015-12-15
fixed role select drop down

eparker71:


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -43,6 +43,10 @@
                 $scope.lookup(searchTerm);
             }
         };
+        $scope.roleselected = function(role_id, role_name){
```
rascalking: style nitpicks:
* js is supposed to prefer camelCase, right?
* the other methods are of the form verbNoun(), can we stick with that?

rascalking:
:+1:

course_info/static/course_info/js/controllers/PeopleController.js
eparker71: yup, that's my bad. typo.

rascalking: Heh, I meant the params as well as the method name.  Don't sweat it, I'll tweak it while I'm doing the user lookup.

PR #31, opened 2015-12-18
Sprint 38 TLT-1155 Fix alert for canvas enrollment failure

rascalking:
Only show the warning about Canvas enrollment failure if we know the course has external sites.

* also, remove unnecessary `console.log`


eparker71:
:+1:

PR #32, opened 2015-12-18
Task/eparker71/tlt 1155 sub branch fix some ui items

eparker71:
addressed how the add user button becomes disabled and available depending on the user search.
also enabled the progress bar when a search is in progress.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -112,21 +112,6 @@
         };
         $scope.disableAddUserButton = function(){
 
-            /*
-            * this method will disable the button when a search is in progress
-            * it will enable the button if multiple results are found and one
-            * is selected.
-            * */
-            //if ($scope.searchTerm.length > 0 && ! $scope.searchInProgress){
-            //    return false;
-            //}
-            //
-            //if( $scope.searchResults.length > 1 ){
-            //    return (!$scope.selectedResult.id);
-            //}
-            //
-            //return true;
-
             if ($scope.searchInProgress) {
                 return true;
             }
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -11,13 +11,14 @@
             2: 'role__role_name',
             3: 'source_manual_registrar',
         };
-
+        $scope.searchInProgress = false;
         // set up functions we'll be calling later
         $scope.addUser = function(searchTerm) {
+            $scope.searchInProgress = true;
             if ($scope.searchResults.length > 1 && $scope.selectedResult.id) {
                 // TODO: assert that radio box is selected
                 $scope.addUserToCourse(searchTerm,
-                                       {user_id: filteredResults[0].univ_id,
+                                       {user_id: $scope.selectedResult.id,
                                         role_id: $scope.selectedRole.roleId});
             }
             else if ($scope.searchResults === 1) {
@@ -69,6 +70,7 @@
                         roleName: $scope.getRoleName(user.role_id),
                     });
                 });
+            $scope.searchInProgress = false;
         };
         $scope.closeAlert = function(source, index) {
             $scope[source].splice(index, 1);
@@ -91,15 +93,22 @@
                 : b.active > a.active;
         };
         $scope.disableAddUserButton = function(){
-            if ($scope.searchResults.length > 0 ){
-                return (!$scope.selectedResult.id);
-            }
-            else if ($scope.searchTerm.length > 0){
+
+            /*
+            * this method will disable the button when a search is in progress
+            * it will enable the button if multiple results are found and one
+            * is selected.
+            * */
+            if ($scope.searchTerm.length > 0 && ! $scope.searchInProgress){
                 return false;
             }
-            else {
-                return true;
+
+            if( $scope.searchResults.length > 1 ){
+                return (!$scope.selectedResult.id);
             }
+
+            return true;
+
         };
         $scope.filterResults = function(searchResults){
             var filteredResults = Array();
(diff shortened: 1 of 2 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -11,13 +11,14 @@
             2: 'role__role_name',
             3: 'source_manual_registrar',
         };
-
+        $scope.searchInProgress = false;
```
rascalking: Can we move this down to live with the other "initializing the scope" entries after the function definitions?

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -83,6 +84,7 @@
                         roleName: $scope.getRoleName(user.role_id),
                     });
                 });
+            $scope.searchInProgress = false;
```
rascalking: Don't we want to wait to set this to false until the http calls we triggered are finished?

course_info/templates/course_info/partials/people.html
```diff
@@ -146,7 +146,8 @@
 								</a>
 							</div>
 
-              <div class="col-md-8 col-md-offset-2 hidden" style="padding-top: 1em">
+              <div class="col-md-8 col-md-offset-2"
+                   style="padding-top: 1em" ng-class="{'hidden' : !searchInProgress || searchResults.length > 0}">
```
rascalking: Doesn't this need `{{ }}`, not `{ }`?

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -105,15 +107,22 @@
...
-                return (!$scope.selectedResult.id);
-            }
-            else if ($scope.searchTerm.length > 0){
+
+            /*
+            * this method will disable the button when a search is in progress
+            * it will enable the button if multiple results are found and one
+            * is selected.
+            * */
+            if ($scope.searchTerm.length > 0 && ! $scope.searchInProgress){
```
rascalking: I think you can just add a check for search in progress at the beginning to the previous logic.
```
if ($scope.searchInProgress) {
    return true;
}
else if ($scope.searchResults.length > 0) {
    return (!$scope.selectedResult.id);
}
else if ($scope.searchTerm.length > 0) {
    return false;
}
else {
    return true;
}
```

course_info/templates/course_info/partials/people.html
eparker71: not according to the docs, { 'class' : test } is how they have it

course_info/static/course_info/js/controllers/PeopleController.js
eparker71: sure

course_info/static/course_info/js/controllers/PeopleController.js
eparker71: so, that works, but I swear it didn't before argh!!! :)

course_info/static/course_info/js/controllers/PeopleController.js
eparker71: for this, are you thinking use the $q for the post and the get? like you have on line 258 
```           $q.all([peoplePromise, memberPromise])
                .then($scope.handleLookupResults);```

File: course_info/templates/course_info/partials/people.html (hunk 3 above)
Comment (rascalking) [inline course_info/templates/course_info/partials/people.html]: Ahhh, it's a map of classes to expressions that determine whether that class is active or not.  Clever.

File: course_info/static/course_info/js/controllers/PeopleController.js (hunk 4 above)
Comment (rascalking) [inline course_info/static/course_info/js/controllers/PeopleController.js]: As long as it works now.  :)

File: course_info/static/course_info/js/controllers/PeopleController.js (hunk 2 above)
Comment (rascalking) [inline course_info/static/course_info/js/controllers/PeopleController.js]: Nope, I was just thinking that wherever we know we're done with network calls, we set `$scope.searchInProgress = false`.  Using `$q` might not work...the GET only happens if the POST succeeds, right?

Comment (eparker71) [inline course_info/static/course_info/js/controllers/PeopleController.js]: yeah, looking at the possible ways the network call can end, I think we want to set this when either the get is successful or fails (post succeeded), or the post fails? sound right?

Comment (rascalking) [inline course_info/static/course_info/js/controllers/PeopleController.js]: yup.  i think you can do that by chaining a .then() on the end of the get, and putting the call inside the failure callback on the post.

Comment (eparker71) [inline course_info/static/course_info/js/controllers/PeopleController.js]: i was looking for a finally or always but angular doens't seem to have any. chaining then will do it though :)

File: course_info/static/course_info/js/controllers/PeopleController.js [hunk 5]
```diff
@@ -62,6 +62,7 @@
                             data.results[0].searchTerm = searchTerm;
                             $scope.successes.push(data.results[0]);
                             $scope.dtInstance.reloadData();
+
```
rascalking: no need to add an empty line here.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -82,7 +86,9 @@
                         searchTerm: searchTerm,
                         roleName: $scope.getRoleName(user.role_id),
                     });
+                    $scope.searchInProgress = false;
                 });
+
```
rascalking: again, no need for the empty line

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -105,15 +111,20 @@
                 : b.active > a.active;
         };
         $scope.disableAddUserButton = function(){
-            if ($scope.searchResults.length > 0 ){
+
```
rascalking: ditto

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -105,15 +111,20 @@
...
                 return (!$scope.selectedResult.id);
             }
-            else if ($scope.searchTerm.length > 0){
+            else if ($scope.searchTerm.length > 0) {
                 return false;
             }
             else {
                 return true;
             }
+
```
rascalking: ditto

rascalking:
Just whitespace whining.  :+1:

PR #34, opened 2015-12-21
Sprint 38 TLT-1155 Add cancel button for choose-from-multiple-roles prompt

rascalking:
behind the scenes, it just clears out the searchResults.  make sure we
do the same thing when the person adds a specific role, don't force them
to then cancel out by hand.


eparker71:
:+1:

PR #36, opened 2016-01-05
Sprint 38 TLT-1155 Jasmine tests

rascalking:
Includes changes to the controller itself:
* Clean up the logic on what to do when the add user button is pressed, based on the state in the scope
* Removes the special casing of canvas partial failures, as the API is no longer attempting to enroll in a canvas site we don't think should exist.
* Use the angular $log service for logging, to facilitate mocking during testing.  (this also makes it easier to log to the server down the road)
* Fix a bug where we weren't clearing the search results and turning off the searchInProgress flag when we got an error retrieving the enrollment details after an apparently successful user add.
* Fix the compareRoles method, which returned true/false (ie. equal or not) when it needed to return -1/0/1 (ie. less than/equal/greater than).
* Sorted the scope variable declarations.

Testing changes:
* Added a new library requirement to package.json/karma.conf.
* Fixed existing tests.
* Added tests for all the new methods added to the PeopleController.


course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -1,13 +1,44 @@
+// regex taken from https://www.ietf.org/rfc/rfc3986.txt appendix B, which
```
elliottyates: That's pretty old, although of course the def of a URI is also old. :) Is there any way to tell if there's another more recent RFC or accepted recommendation that builds on or fixes any bugs in this regex?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -24,8 +56,9 @@ describe('Unit testing PeopleController', function() {
             };
         });
         scope = $rootScope.$new();
-        $routeParams.course_instance_id = 1234567890;
+        $routeParams.courseInstanceId = 1234567890;
```
elliottyates: This is picky, but could this and the `courseInstanceURL` var defined above be derived from the same var somehow? Or is that more trouble/ambiguous than it's worth to make it DRYer?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -35,42 +68,41 @@ describe('Unit testing PeopleController', function() {
...
+        beforeEach(function() {
+            controller = $controller('PeopleController', {$scope: scope });
+        });
+
         afterEach(function() {
             // handle the course instance get from setTitle, so we can always
             // assert at the end of a test that there's no pending http calls.
-            $httpBackend.expectGET('/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args=api%2Fcourse%2Fv2%2Fcourse_instances%2F1234567890%2F')
-                .respond(200, '');
+            $httpBackend.expectGET(courseInstanceURL).respond(200, '');
```
elliottyates: Just so I understand, this says "the next HTTP request in the async stack is this 200, and I'm going to flush just that one, so that when the parent's `afterEach()` checks `verifyNoOutstandingExpectation()` it will find nothing left to flush?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+                    {active: 1, prime_role_indicator: ''},
+                    {active: 0, prime_role_indicator: 'N'},
+                    {active: 1, prime_role_indicator: 'N'},
+                    {active: 0, prime_role_indicator: 'Y'},
+                    {active: 1, prime_role_indicator: 'Y'},
+                ];
+                var afterSorting = [
+                    {active: 1, prime_role_indicator: 'Y'},
+                    {active: 1, prime_role_indicator: 'N'},
+                    {active: 1, prime_role_indicator: ''},
```
elliottyates: Looking at this now, I wonder if perhaps `''` should be ranked higher than `'N'`... maybe a `''` is a sort of implicit `'Y'` while `'N'` is more explicitly not `'Y'`... anyways. Probably there's not a soul at Harvard that can address that with confidence. :)

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+            // the state of the add user button depends on the contents of
+            // the search field, on whether a lookup has returned multiple
+            // results, and on whether any of those results have been selected
+            it('should disable if a search is in progress', function() {
+                scope.searchInProgress = true;
+                expect(scope.disableAddUserButton()).toBe(true);
+            });
+            it('should enable if the search had multiple results and one is selected',
+               function() {
+                   scope.searchResults = [{}]; // contents don't matter, only length
```
elliottyates: I see in PeopleController.js that `$scope.searchResults` is set iff there are multiples; it's a little confusing though that 'multiple results' as described in the `it()` clause is equivalent to a single-item array, since there's not a clear indication by the scope variable name that it will only be populated if there are multiple results. Could these maybe be `[{}, {}]` just to be a little more life-like?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+                var searchResults = [
+                    {univ_id: 123, active: 0, prime_role_indicator: ''},
+                    {univ_id: 123, active: 1, prime_role_indicator: 'Y'},
+                    {univ_id: 456, active: 0, prime_role_indicator: 'Y'},
+                    {univ_id: 456, active: 1, prime_role_indicator: ''},
+                    {univ_id: 789, active: 0, prime_role_indicator: 'Y'},
+                ];
+                var filtered = scope.filterResults(searchResults);
+                var uniq = {};
+                filtered.forEach(function(r) { uniq[r.univ_id] = true; });
```
elliottyates: Won't this just overwrite an existing key, so there will only ever be one value per univ_id anyway? Maybe we should be doing `uniq[r.univ_id] = (r.univ_id in uniq ? uniq[r.univ_id]++ : 1)` or something like that, and checking both `.length()` and that all values are `1`?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+            });
+        });
+
+        describe('handleAjaxError', function() {
+            it('should log an error', function() {
+                var expectedMessage = 
+                    "Error getting data from https://tea.pot: 418 I'm a teapot: {}";
+                scope.handleAjaxError({}, 418, {}, {url: 'https://tea.pot'},
+                                      "I'm a teapot");
+                expect($log.error.logs).toEqual([[expectedMessage]]);
```
elliottyates: I wonder why the `$log.x.logs` array stores individual log entries as arrays... strange. :|

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+        it('should call lookup if there are no search results', function() {
+            scope.searchResults = [];
+            scope.addUser('bob');
+            expect(scope.addUserToCourse.calls.count()).toEqual(0);
+            expect(scope.lookup.calls.count()).toEqual(1);
+            expect(scope.lookup.calls.argsFor(0)).toEqual(['bob']);
+        });
+
+        it('should log an error if called with a single search result', function() {
+            scope.searchResults = [{}]; // contents don't matter
```
elliottyates: (referencing above comment re: making test data more 'real-world'-like, here, 'single search result' is represented by `[{}]` (because the number actually matters when calling `addUser()`, whereas above it's conflated with multiple results.)

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+               scope.selectedRole = {roleId: 888};
+               scope.addUser('bob');
+               expect(scope.addUserToCourse.calls.count()).toEqual(1);
+               expect(scope.addUserToCourse.calls.argsFor(0)).toEqual(
+                       ['bob', {user_id: scope.selectedResult.id,
+                                role_id: scope.selectedRole.roleId}]);
+           }
+        );
+    });
+    describe('addUserToCourse', function() {
```
elliottyates: space

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+            expect(scope.dtInstance.reloadData.calls.count()).toEqual(1);
+        });
+
+        it('should alert on a partial failure', function() {
+            var partialFailureResponse = 
+                JSON.parse(JSON.stringify(enrollmentDetails.results[0]));
+            partialFailureResponse.detail = 'Some kind of partial failure';
+            var expectedSuccess = 
+                JSON.parse(JSON.stringify(enrollmentDetails.results[0]));
+            expectedSuccess.searchTerm = searchTerm;
```
elliottyates: This occurs in a couple of `it()`s, can it be in the `beforeEach()` or at the `describe()` level?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+            var expectedPartialFailure = {
+                searchTerm: searchTerm,
+                text: partialFailureResponse.detail,
+            };
+
+            // mock out the datatable so we can verify that it gets reloaded
```
elliottyates: ditto -- this occurs in a couple of `it()`s, can it be in the `beforeEach()` or at the `describe()` level?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+            $httpBackend.flush(2);
+
+            // check to see if it's reacting correctly
+            expect(scope.successes).toEqual([expectedSuccess]);
+            expect(scope.partialFailures).toEqual([expectedPartialFailure]);
+            expect(scope.dtInstance.reloadData.calls.count()).toEqual(1);
+        });
+
+        it('should warn on failure to add the user', function() {
+            spyOn(scope, 'handleAjaxError');
```
elliottyates: ditto -- this occurs in a couple of `it()`s, can it be in the `beforeEach()` or at the `describe()` level?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+                           .respond(201, JSON.stringify(user)); 
+               $httpBackend.expectGET(enrollmentDetailsURL).respond(404, '');
+               $httpBackend.flush(2);
+
+               expect(scope.partialFailures).toEqual([expectedPartialFailure]);
+               expect(scope.handleAjaxError.calls.count()).toEqual(1);
+           }
+        );
+    });
+    describe('handleLookupResults', function() {
```
elliottyates: space

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+                title: 'handleLookupResults test',
+            };
+            courseInstances.instances[ci.course_instance_id] = ci;
+            controller = $controller('PeopleController', {$scope: scope});
+            scope.searchInProgress = true;
+        });
+
+        it('should warn and disable progress if the user is already enrolled',
+           function() {
+               var peopleResult = {};
```
elliottyates: maybe a small comment like you've done above that this var/content doesn't matter would be helpful.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -128,19 +160,540 @@ describe('Unit testing PeopleController', function() {
...
+               var memberResult = {
+                   data: {results:
+                              [{profile: {name_last: 'Dobbs',
+                                          name_first: 'Bob'}}]},
+                   config: {searchTerm: 'bob_dobbs@harvard.edu'},
+               };
+               var expectedWarning = {
+                   type: 'alreadyInCourse',
+                   fullName: 'Dobbs, Bob',
+                   memberships: memberResult.data.results.slice(),
```
elliottyates: Why `slice()`? Wouldn't `results` suffice?

elliottyates:
:+1: and thanks for the useful comments

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Short answer: This is only used to test a url generated by our testing, so as long as it works, I'm happy with it.  If this were user-facing code, I'd be paranoid enough to pull in a utility library that has some test coverage to back it up.

Long answer: Per https://www.rfc-editor.org/info/rfc3986, there are two "updated by" RFCs, one referencing ipv6 (which the regex will fail on), and one offering suggestions on restricting uri definitions based on the scheme (we only care about http/https).  There is an erratum correcting the regex syntax, including escaping, so I've switched to using that one.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Sure.  I suspect it's a wash either way, but not a hard change to make.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: It says "expect (assert) that the next call to the mock `$http` is to the `courseInstanceURL`, and return a 200 with '' for the body when it gets flushed".  The next line explicitly flushes just one pending http call.

And yeah, we're doing this so that we can make those `verifyNoOutstandingExpectation()`/`verifyNoOutstandingRequest()` calls and have them pass.  The former of which makes sure that all the `expect*()` calls have fired, the latter of which makes sure all the `$http` calls have fired.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yeah, I'm not touching that logic bundle with a 10 foot pole again, except at gunpoint.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Sure.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yes, we should totally be doing that.  Good catch!

rascalking: Actually, no, we don't need to do that.  :)

Clearly I need some comments here.  If there are duplicates, then the `filteredIds` equality check will pass, but the `filtered.length` check won't.  The only way both pass is if we get back exactly 3 records, each of which has a unique id, and those ids match.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: I *think* it's because `log.error()` et al. are variadic, so even though we only pass in one argument, it gets an array.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yup.  Fixed the one above.  Probably a way to refactor the controller to have that make more sense.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: thanks for the catch

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: I usually go with a threshold of 3 tests to bundle things into a sub-describe in order to pull things into `beforeEach`/`afterEach`.  The extra indent you give up in order to sub-describe is enough of a pain for me to not bother with 2.

But you're right, this could go into the `beforeEach` that's a peer to the `it` calls.

rascalking: Hrm...when I went to do this, I noticed a bunch of other `expectedX`-type definitions that I could move up as well.  Moving them all away means the test data is further away from the test code than I'd like, and it feels like it hurts readability.  Think I'll pass on changing this, let me know if you feel strongly about it.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yes, but it only occurs in two of the four tests, and I'd rather scope the mocks to just those tests where they're needed.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Same objection to mocking where it's not necessary as above.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: yup

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: done

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
rascalking: Yeah...I was feeling paranoid about the method messing with the result as passed in.  But now that I think of it, `slice()` isn't really doing what I want it to...it gets me a new array with a new reference to the same underlying `{profile: ...}` object, so if the callee mutates that object, I'm not protected.

I can use JSON as a hack to get a deep copy, but that's even more awkward.  I'll just remove the `slice()`.

PR #38, opened 2016-01-05
Task/josie/tlt 1155 selenium

sapnamysore:
This PR contains selenium tests for 1155 and also some refactoring and changes per the best practices being discussed in general for selenium.


canvas_account_admin_tools/settings/local.py
```diff
@@ -21,11 +21,34 @@
...
+         'existing': {
+            'user_id': '20299916'
+         },
+         'new': {
+            'role': 'Teacher',
+            'user_id': '30833767'
+         }
+      }
+   },
+   'run_locally': True,
```
rascalking: This probably wants to be `SECURE_SETTINGS.get('selenium_run_locally', False)`, to match what we're doing in other projects.

course_info/templates/course_info/partials/people.html
```diff
@@ -180,7 +180,7 @@
 
       <h3>People in {{title}}</h3>
 
-      <uib-alert ng-repeat="person in successes" type="success"
+      <uib-alert id="success-alert" ng-repeat="person in successes" type="success"
```
rascalking: This will create multiple alerts, one for each item in `$scope.successes`, so we need to add a number or something else to the id here.  `id="success-alert-{{$index}}"` will probably work, but that `$index` will track the alert's position in `$scope.successes`.  That is to say, if you remove `success-alert-0`, the ones after it will all see their id change.

selenium_tests/course_info/course_info_base_test_case.py
```diff
@@ -4,32 +4,54 @@
...
+from selenium_tests.course_info.page_objects.course_info_search_page_object import \
+    CourseSearchPageObject
 
 
 class CourseInfoBaseTestCase(BaseSeleniumTestCase):
-    """
-    Bulk Create base test case, all other tests will subclass this class
-    """
+
+    _search_page = None
```
rascalking: This seems to be used as an instance member, not a class member, I wouldn't define it up here.

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -0,0 +1,37 @@
...
+            year=course['year'], search_term=course['cid'])
+
+        # click on course link to view list of people in course
+        search_page.select_course(cid=course['cid'])
+
+        # search for a user and add user to course
+        people_page.search_and_add_user(new_user['user_id'], new_user['role'])
+
+        # assert that user is found
+        self.assertTrue(people_page.is_person_on_page(new_user['user_id']))
```
rascalking: This will only be true if the course has few enough users that they all fit on one page.  Worth commenting on that here in case the data changes.

canvas_account_admin_tools/settings/local.py
sapnamysore: yeah, will fix it

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -0,0 +1,65 @@
...
+    ADD_PEOPLE_SEARCH_TXT = (By.ID, "emailHUID")
+    ADD_TO_COURSE_BUTTON = (By.ID, "add-user-btn-id")
+    ALERT_SUCCESS_PERSON = (By.XPATH, '//p[contains(.,"just added")]')
+    ROLES_DROPDOWN_LIST = (By.ID, "select-role-btn-id")
+
+    @classmethod
+    def TD_TEXT_XPATH(cls, search_text):
+        """ returns a locator for a table cell element in the people table;
+        search_text should be user's name, user_id, etc """
+        return By.XPATH, '//td[contains(text(), "{}")]'.format(search_text)
```
rascalking: My XPATH-fu is terrible.  If there's a way to scope this to be sure we're within the datatable, though, this would be a lot less fragile.  As is, we'll match any <TD> on the page with the search text in it, right?

rascalking:
The add-success alert id has to be fixed unless we can guarantee no more than one success alerts on the page at once.  Everything else could go in as-is, and we could clean it up later.

selenium_tests/course_info/course_info_add_people_tests.py
sapnamysore: True. Followed this up with asserting for the 'successfully added' message.  Added the comment as suggested.

course_info/templates/course_info/partials/people.html
sapnamysore: Good catch. Turns out we are not using this id anymore. Removing it

PR #39, opened 2016-01-06
Sprint 39 TLT-2299 Don't couple the search bar and the success alert

rascalking:
We were just referencing the scope variable, instead of the copy from the success entry.


elliottyates:
:+1:

PR #2, opened 2015-10-01
Sprint 32 TLT-2038 Proxy icommons_rest_api requests

douglashall:
* Use django-proxy to proxy requests to icommons_rest_api with an Authentication token
* This allows requests to be made to this application with the following url structure and those requests will be forwarded on to the configured icommons_rest_api deployment:
```
[canvas_account_admin_tools host]/icommons_rest_api/api/course/v2/course_instances?format=json&canvas_course_id=6402
```

PR #41, opened 2016-01-11
Sprint 39 TLT-1997 Delete functionality for course detail page

rascalking:
NOTE:
* Fixes to the jasmine unit tests are coming in a follow-on PR.
* This currently depends on changes that live on the `task/elliottyates/tlt-1997/remove_people` branch of icommons_rest_api.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -119,6 +119,8 @@
             }
         };
         $scope.confirmRemove = function(membership) {
+            // creates a new remove user confirmation modal, and
+            // stashes this membership object on the modal's child scope.
             var modalInstance = $uibModal.open({
                 controller: function($scope, membership) {
                     $scope.membership = membership;
@@ -130,6 +132,8 @@
                 },
                 templateUrl: 'partials/remove-course-membership-confirmation.html',
             });
+
+            // if they confirm, then do the work
             modalInstance.result.then($scope.removeMembership);
         };
         $scope.disableAddUserButton = function(){
@@ -173,12 +177,6 @@
         $scope.handleAjaxError = function(data, status, headers, config, statusText) {
             $log.error('Error getting data from ' + config.url + ': ' + status +
                        ' ' + statusText + ': ' + JSON.stringify(data));
-        };
-        $scope.handleRemoveResults = function(membership) {
-            var success = membership; // TODO - copy to avoid stomping the original?
-            success.searchTerm = membership.user_id;
-            success.action = 'removed from';
-            $scope.successes.push(success);
         };
         $scope.handleLookupResults = function(results) {
             var peopleResult = results[0];
@@ -278,16 +276,22 @@
                                       ['api/course/v2/course_instances/' +
                                        $scope.courseInstanceId +
                                        '/people/' + membership.user_id]);
-            var params = {
-                role_id: membership.role.role_id,
-                user_id: membership.user_id,
+            var config = {
+                data: {
+                    role_id: membership.role.role_id,
+                    user_id: membership.user_id,
+                },
+                headers: {
+                    'Content-Type': 'application/json',
+                },
             };
-            $scope.handleRemoveResults(membership);
-
-            return;
-            $http.delete(courseMemberURL, {params: params})
+            $http.delete(courseMemberURL, config)
                  .success(function() {
-                     $scope.handleRemoveResults(membership);
+                     var success = membership; // TODO - copy to avoid stomping the original?
+                     success.searchTerm = membership.user_id;
+                     success.action = 'removed from';
+                     $scope.successes.push(success);
+                     $scope.dtInstance.reloadData()
                  })
                  .error($scope.handleAjaxError) // TODO - real error handling
                  ;
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -175,8 +175,8 @@
             return filteredResults;
         };
         $scope.handleAjaxError = function(data, status, headers, config, statusText) {
-            $log.error('Error getting data from ' + config.url + ': ' + status +
-                       ' ' + statusText + ': ' + JSON.stringify(data));
+            $log.error('Error attempting to ' + config.method + ' ' + config.url +
+                       ': ' + status + ' ' + statusText + ': ' + JSON.stringify(data));
         };
         $scope.handleLookupResults = function(results) {
             var peopleResult = results[0];
@@ -281,9 +281,6 @@
                     role_id: membership.role.role_id,
                     user_id: membership.user_id,
                 },
-                headers: {
-                    'Content-Type': 'application/json',
-                },
             };
             $http.delete(courseMemberURL, config)
                  .success(function() {
@@ -311,7 +308,8 @@
             var cell = '<div class="text-center">';
             if (!registrarFed) {
                 cell += '<a href="" ng-click="confirmRemove(' +
-                        'dtInstance.DataTable.data()[' + meta.row + '])">';
+                        'dtInstance.DataTable.data()[' + meta.row + '])" ' +
+                        'data-sisid="' + full.user_id + '">';
             }
             cell += '<i class="fa fa-trash-o"';
             if (registrarFed) {
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -1,3 +1,4 @@
+// TODO - split out add/remove into separate controllers?
 (function() {
     var app = angular.module('CourseInfo');
     app.controller('PeopleController', PeopleController);
@@ -50,7 +51,7 @@
                         // log it, then display a warning
                         $scope.handleAjaxError(data, status, headers, config,
                                 statusText);
-                        $scope.partialFailures.push({
+                        $scope.addPartialFailures.push({
                             searchTerm: searchTerm,
                             text: 'Add to course seemed to succeed, but ' +
                                 'we received an error trying to retrieve ' +
@@ -72,14 +73,14 @@
                             (data.detail.indexOf('Canvas API error details') != -1)) {
                         // partial success, where we enrolled in the coursemanager
                         // db, but got an error trying to enroll in canvas
-                        $scope.partialFailures.push({
+                        $scope.addPartialFailures.push({
                             searchTerm: searchTerm,
                             text: data.detail,
                         });
                         handlePostSuccess();
                     }
                     else {
-                        $scope.warnings.push({
+                        $scope.addWarnings.push({
                             type: 'addFailed',
                             searchTerm: searchTerm,
                         });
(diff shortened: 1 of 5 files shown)


elliottyates:
@elliott

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -248,6 +268,73 @@
...
+                    $scope.handleAjaxError(data, status, headers, config, statusText);
+
+                    var failure = membership; // TODO - copy to avoid stomping the original?
+                    var reloadData = false; // for some partial failures we need to reload
+                    switch(status) {
+                        case 404:
+                            switch (data.detail) {
+                                case 'User not found.':
+                                    failure.type = 'noSuchUser';
+                                    break;
```
elliottyates: Seems like we should probably refresh the datatable here as well so that person disappears from the visible results.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -255,6 +342,28 @@
         $scope.renderName = function(data, type, full, meta) {
             return full.profile.name_last + ', ' + full.profile.name_first;
         };
+        $scope.renderRemove = function(data, type, full, meta) {
+            // TODO - maybe make this a directive?  the isolate scope
+            //        in a directive complicates things.  anything has
+            //        to be better than this mess, though.
```
elliottyates: Although it would require more actual concatenations, it might read easier if it was built in a nested way, not a linear way. Not sure if that would work for all cases, though.

e.g.:
```
cellStyle = registrarFed ? ' style="color: #ccc;"' : '';
cellLinkOpening = !registrarFed ? .... etc
...
cell = '<i class="fa fa-trash-o"' + cellStyle + '></i>';
cell = cellLinkOpening + cell + cellLinkClosing;
cell = '<div ...' + cell + '</div>';
return cell;
```

Also since we're using `registrarFed` logic in multiple places would it make sense to make it a computed property of the row data and refer to that instead of re-running the regex each time? Won't save much time with only 10 records being displayed, I suppose. But we have sorting on the Manually/Registrar added field, so seems like we should be able to just reference that somehow.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -255,6 +342,28 @@
...
+            var registrarFed = /^.*feed$/.test(full.source);
+            var cell = '<div class="text-center">';
+            if (!registrarFed) {
+                cell += '<a href="" ng-click="confirmRemove(' +
+                        'dtInstance.DataTable.data()[' + meta.row + '])" ' +
+                        'data-sisid="' + full.user_id + '">';
+            }
+            cell += '<i class="fa fa-trash-o"';
+            if (registrarFed) {
+                cell += ' style="color: #ccc;"';
```
elliottyates: is an inline style what we want here vs class-based styling?

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -383,6 +493,12 @@
                 render: $scope.renderSource,
                 title: 'Source',
             },
+            {
+                data: '',
+                render: $scope.renderRemove,
+                title: 'Remove',
+                // TODO - flag as not sortable
```
elliottyates: Yes, this is one of my comments - remove column shouldn't be sortable.

course_info/templates/course_info/partials/remove-course-membership-confirmation.html
```diff
@@ -0,0 +1,28 @@
...
+<div class="modal-header">
+	<button type="button" class="close" ng-click="$dismiss('cancel')"
+          id="modalClose">
+    &times;
+  </button>
+	<h4 class="modal-title">Confirm Remove User</h4>
+</div>
+<div class="modal-body">
+  <p>Are you sure you want to remove</p>
+  <div style="border-style: solid; border-width: 1px 0px 1px 0px; border-color: #ccc; padding:.5em; margin-top:1em; margin-bottom:1em;">
```
elliottyates: is an inline style what we want here vs class-based styling?

elliottyates:
(FYI still working on the large template change)

course_info/templates/course_info/partials/people.html
```diff
@@ -10,200 +10,247 @@
...
+          to find someone's official email address.
+        </div>
 
-              <div ng-if="warning.type == 'addFailed'">
-                <p><strong>{{ warning.searchTerm }}</strong> could not be added to this course at this time.</p>
-                <p>Please try again.</p>
+        <!-- begin add warning alerts -->
+        <uib-alert ng-repeat="warning in addWarnings" type="warning"
+                   close="closeAlert('addWarnings', $index)" class="col-md-12">
+          <div ng-switch on="warning.type">
```
elliottyates: Seems like it would be nice for this to be a directive, or a partial of some sort, along with other big chunks of if/switch-style display code.

course_info/templates/course_info/partials/people.html
```diff
@@ -10,200 +10,247 @@
...
+            <!-- user was already enrolled -->
+            <div ng-switch-when="alreadyInCourse">
+              <p><strong>{{ warning.searchTerm }}</strong> was found in the course:</p>
+              <p>{{ warning.fullName }} has the following email/id and role associated with this course:</p>
+              <div ng-repeat="membership in warning.memberships"
+                   style="border-style: solid; border-width: 1px 0px 1px 0px; border-color: #ccc; padding:.5em; margin-top:1em;">
```
elliottyates: Here and elsewhere there a quite a few inline styles, should we be abstracting these with classes?

course_info/templates/course_info/partials/people.html
```diff
@@ -10,200 +10,247 @@
...
+    <div ng-switch on="failure.type">
+      <div ng-switch-when="noSuchUser">
+        <strong>{{ failure.profile.name_last }}, {{ failure.profile.name_first }}</strong>
+        has already been removed from this course.
+      </div>
+      <div ng-switch-when="noSuchCourse">
+        <strong>{{ failure.profile.name_last }}, {{ failure.profile.name_first }}</strong>
+        cannot be removed from this course because it no longer exists.
+      </div>
+      <div ng-switch-when="serverError">
```
elliottyates: `serverError` and `canvasError` could fall through to `ng-switch-default` to save space, but this is certainly more explicit if you prefer that.

elliottyates:
This is unrelated to delete functionality, but while testing I noticed something... I would expect that Add to Course, when successful, should clear any yellow messages in the Add to Course area, like 'not found' and 'is already in the course' (perhaps not green success messages though).

elliottyates:
Could Remove column be thinner (minimum width) to match the UI mockups more closely?

elliottyates:
The delete confirmation message doesn't match [UI mockup screen #2](https://wiki.harvard.edu/confluence/pages/viewpage.action?title=Admin+Tools+-+Add+People+-+TLT-1997&spaceKey=UIS), but I think it looks good, so maybe UI wiki page just hasn't been updated.

elliottyates:
:+1: I think some refactoring to move view logic into partials for readability would be helpful, and possibly we should try to move CSS styling behind class definitions, but otherwise looks good. I'd say one thing that would be good to fix before this release would be the sort on the remove column.

elliottyates:
Ah, I also noticed (unrelated to this change package) that the add to course button doesn't have the tab index behavior I'd expect -- it gets skipped after the privacy disclaimer and you go to the datatable instead.

course_info/static/course_info/js/controllers/PeopleController.js
rascalking: Good point.

course_info/static/course_info/js/controllers/PeopleController.js
rascalking: probably not...i've been pretty lazy about leaving the styles from vittorio's sample code as inline.

course_info/static/course_info/js/controllers/PeopleController.js
rascalking: thanks for the catch, forgot to go back and figure that out.

course_info/templates/course_info/partials/remove-course-membership-confirmation.html
rascalking: nope.  that same style is used in a few other places...i think there's one spot where it varies, maybe in the modal partial?

course_info/templates/course_info/partials/people.html
rascalking: yup.

course_info/templates/course_info/partials/people.html
rascalking: they've all got slightly different error messages, not sure how fall-through would work.

elliottyates: Yep, got it, that's me not reading carefully.

course_info/static/course_info/js/controllers/PeopleController.js
rascalking: I like the approach you outlined above for cleaning this up...makes it much more readable.

The row data is being pulled by datatables as needed...not sure there's good hooks for adding to the prototype there.  And this is called as a peer to the `renderSource()` method, so I don't think we can safely assume that the rendered cell is going to be available for us to inspect.  Anything I come up with to avoid running that regex twice is more complicated, and way less readable, than just running it twice.

course_info/templates/course_info/partials/people.html
rascalking: Eh, it's not big enough trigger my personal "yank it out" instinct, and it makes it clear from looking at a single file the different alerts that are possible.  I can more easily see the argument for making each type of alert its own partial, and just `ng-include`ing them from the switch statement here.  Not opposed to pulling things out, though, if you feel strongly about it.

rascalking:
re: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/41#issuecomment-170651750

Sure, I can see that making sense.  There's probably other edge cases around when/if to automatically clear these...I'd like to talk it over with Vittorio before implementing it.

rascalking:
re: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/41#issuecomment-170652001

Possibly...I don't know what kind of control datatables gives us over that.

rascalking:
re: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/41#issuecomment-170652152

That was another one where I would normally have gone over it with Vittorio, but yeah...in previous conversations, we'd stressed keeping that row-looking div as a standard indicator of an enrollment in the alerts.  Abandoning it in the modal seemed like an oversight.  I didn't update the UI spec because I wanted to talk it through with him first.  Probably should have at least put an explainer in the ticket.

rascalking:
re: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/41#issuecomment-170653031

Yup, that bothers me, too.  Tech debt it?

elliottyates:
re: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/41#issuecomment-170661849

Sure: https://jira.huit.harvard.edu/browse/TLT-2324

PR #47, opened 2016-01-12
fixed for getting school list where user cannot see sis_account_id

eparker71:
get a list of all the accounts
get a list of the accounts the user can see (they won't contain the sis_account_id)
use the user list to filter out the accounts from the all accounts list so it only contains the accounts the user has access to.

### DIFF
--- a/course_info/canvas.py
+++ b/course_info/canvas.py
@@ -23,6 +23,18 @@ def get_administered_school_accounts(canvas_user_id, allowed_roles=ADMINISTRATOR
     cache_key = CACHE_KEY_ACCOUNTS_BY_USER.format(canvas_user_id)
     school_accounts = cache.get(cache_key)
     if school_accounts is None:
+
+        # get all accounts
+        all_canvas_accounts = get_all_list_data(
+            SDK_CONTEXT,
+            get_sub_accounts_of_account,
+            settings.ICOMMONS_COMMON['CANVAS_ROOT_ACCOUNT_ID'],
+            recursive=True)
+
+        # filter so we only have the school accounts
+        all_school_accounts = {a['id']: a for a in all_canvas_accounts
+                               if (a.get('sis_account_id') or '').startswith('school')}
+
         # retrieve accounts this user is directly associated with
         assigned_accounts = get_all_list_data(SDK_CONTEXT, list_accounts,
                                               as_user_id=canvas_user_id)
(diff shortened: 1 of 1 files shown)


course_info/canvas.py
```diff
@@ -23,6 +23,18 @@ def get_administered_school_accounts(canvas_user_id, allowed_roles=ADMINISTRATOR
...
+        # get all accounts
+        all_canvas_accounts = get_all_list_data(
+            SDK_CONTEXT,
+            get_sub_accounts_of_account,
+            settings.ICOMMONS_COMMON['CANVAS_ROOT_ACCOUNT_ID'],
+            recursive=True)
+
+        # filter so we only have the school accounts
+        all_school_accounts = {a['id']: a for a in all_canvas_accounts
+                               if (a.get('sis_account_id') or '').startswith('school')}
```
rascalking: `(a.get('sis_account_id') or '')` can just be `a.get('sis_account_id, '')`

course_info/canvas.py
```diff
@@ -33,26 +45,15 @@ def get_administered_school_accounts(canvas_user_id, allowed_roles=ADMINISTRATOR
...
-                              SDK_CONTEXT, get_sub_accounts_of_account, id_,
-                              recursive=True, as_user_id=canvas_user_id)
-            all_accounts.update({s['id']: s for s in subaccounts})
-
-        # filter down to just school accounts
-        school_accounts = filter(lambda a: (a.get('sis_account_id') or '').startswith('school'),
-                                 all_accounts.values())
+        # filter out the accounts where the user does not have the proper permissions
+        school_accounts = []
+        for k, v in all_school_accounts.iteritems():
```
rascalking: nitpick: can we use more descriptive vars than `k, v`?  maybe `id_, account`?

course_info/canvas.py
```diff
@@ -33,26 +45,15 @@ def get_administered_school_accounts(canvas_user_id, allowed_roles=ADMINISTRATOR
...
+            if k in allowed_accounts.keys():
```
rascalking: don't need to call `.keys()`...checking `x in some_dict` checks against the keys of `some_dict` anyway.

rascalking:
:+1: 

The comments are all style suggestions, nothing that needs to block things.

course_info/canvas.py
eparker71: yeah, those aren't great ;)

course_info/canvas.py
eparker71: cool, one less call

PR #50, opened 2016-01-20
fixed up default url settings

eparker71:
removed CONCLUDE_COURSES_URL setting from base.py
added to CONCLUDE_COURSES_URL setting aws.py and local.py
removed default url setting from aws.py and converted to use [] instead of .get()
local.py now has the code that used to be in base.py

Jaime, I think this is what we talked about. Let me know. 
@bermudezjd @rascalking


canvas_account_admin_tools/settings/aws.py
```diff
@@ -20,3 +20,5 @@
 SESSION_COOKIE_SECURE = True
 
 dictConfig(LOGGING)
+
+CONCLUDE_COURSES_URL = SECURE_SETTINGS['conclude_courses_url']
```
bermudezjd: I would move this up a bit.  Pretty sure we want the last thing in this settings file to be the `dictConfig` line.

canvas_account_admin_tools/settings/local.py
```diff
@@ -57,3 +57,8 @@
    'selenium_grid_url': SECURE_SETTINGS.get('selenium_grid_url'),
    'use_htmlrunner': SECURE_SETTINGS.get('selenium_use_htmlrunner', True),
 }
+
+CONCLUDE_COURSES_URL = SECURE_SETTINGS.get(
+    'conclude_courses_url',
+    'https://icommons-tools.dev.tlt.harvard.edu/course_conclusion'
+)
```
bermudezjd: This will work, but you might want to discuss whether this could be hardcoded to point to the dev url.  Unit tests should be able to mock out a setting value if needed, but I could see an argument for allowing this to point to localhost when running the app locally.

rascalking:
:+1:, modulo what @bermudezjd said about making sure the `dictConfig(LOGGING)` stays at the bottom (maybe add a comment mentioning it, so we don't do this every time someone edits the file?).

canvas_account_admin_tools/settings/local.py
eparker71: sure, we can discus and move in the future if needed.

PR #51, opened 2016-01-25
Change 'Teaching Fellow' to 'TA' in the UI

eparker71:
This change changes the name of the role in the dropdown list as well as the name of the role in the data table. The change to the data table requires a test that checks to see if the role is 'Teaching Fellow' if it is, it returns 'TA', if not, it just returns the current value. This is really a temp fix until the decision is made about changing the values in the UserRoles table.

Jill and Colin are going to discuss making the change to the role name in the database. The icommons rest api returns the values of the UserRoles table which has the role name 'Teaching Fellow'.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -347,6 +347,12 @@
         $scope.renderName = function(data, type, full, meta) {
             return full.profile.name_last + ', ' + full.profile.name_first;
         };
+        // hotfix added to address TA role name
+        // will be addressed with a database change as soon as
+        // devops determines viability of change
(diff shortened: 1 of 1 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -347,6 +347,12 @@
         $scope.renderName = function(data, type, full, meta) {
             return full.profile.name_last + ', ' + full.profile.name_first;
         };
+        // hotfix added to address TA role name
```
sapnamysore: do we want to  remove  'hotfix' from the comment ? Maybe switch to a TODO?

sapnamysore:
:+1:  Looks good!

PR #52, opened 2016-01-25
Task/eparker71/tlt 2349 fix success and error messages stackup

eparker71:
This PR adds a method called clearAllMessages() to the app. This method is called before any new message is added to any of the message lists. This prevents the messages from stacking in the UI.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -396,8 +396,9 @@
                     .error($scope.handleAjaxError);
             }
         };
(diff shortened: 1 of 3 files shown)


rascalking:
Is it worth simply replacing the lists with a scalar variable?  Then where we currently call to push, just assign, and where we currently pop, just assign `null`?

Also, I think this will break the existing "show success as well as a warning when db add succeeds but canvas add fails" functionality.

eparker71:
updated based on pr comments. changed lists to scalars, updated unit tests.

PR #54, opened 2016-01-26
- Cleaned up of test links in the new dev environment

hktest:
Updated deprecated URLs and tested locally - successfully.


canvas_account_admin_tools/settings/local.py
```diff
@@ -22,12 +22,12 @@
 
 SELENIUM_CONFIG = {
    'account_admin': {
-      'relative_url': 'accounts/1/external_tools/20',
+      'relative_url': 'accounts/10/external_tools/79',  # dev (Admin Console)
```
elliottyates: Comment is very useful.

elliottyates:
:+1:

PR #55, opened 2016-01-26
icommons_common unit tests no longer look for non-existent termtool d…

elliottyates:
…atabase connection and run icommons_common models against the default database


rascalking:
:+1:

PR #63, opened 2016-02-02
Task/eparker71/tlt 2349 fix success and error messages stackup

eparker71:
we need to call clear messages before success to make sure there are no other message boxs in the display but we dont' want to clear the partial failure data if it exists.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -457,7 +457,6 @@
         };
 
         $scope.clearMessages = function(){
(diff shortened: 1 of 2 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -108,6 +110,9 @@
         };
         $scope.closeAlert = function(source) {
             $scope[source] = null;
+            if($scope.partialFailureData){
```
rascalking: nitpick: does this need to be in a conditional, or can we just always set it to `null`?

PR #64, opened 2016-02-10
Task/eparker71/tlt 9999 place holder change teaching fellow to ta

eparker71:
this is a temp fix to change the display text of the role "Teaching Fellow" to "TA"
a more perm solution is being discussed, but will involve talking to the schools. Colin is going to have that discussion with the schools and let Jill know.


sapnamysore:
:+1:

PR #65, opened 2016-02-11
Sprint 40 TLT-2435 Keep the people page in existing tab, widen search inputs

rascalking:
And remove some inlined styling while we're at it.  yay!


elliottyates:
:+1:

PR #66, opened 2016-02-11
Task/hchow/2426 new roles check

hktest:
- Added ddt so we can better check users with multiple roles in admin canvas console (in a excel data format)
- Elliott did a good amount of refactoring. :)  

Huge THANKS to @elliottyates for his work, and also for helping me to debug an issue that was blocking me.


selenium_tests/course_info/course_info_base_test_case.py
```diff
@@ -10,11 +9,20 @@
     import AccountAdminDashboardPage
 from selenium_tests.course_info.page_objects.course_info_search_page_object \
     import CourseSearchPageObject
+from selenium_tests.course_info.page_objects.course_people_page_object \
+    import CoursePeoplePageObject
+
+
+# Common files used for all Manage People test cases
+TEST_USERS_WITH_ROLES = join(dirname(abspath(__file__)),
```
rascalking: From the variable name, I'd expect this to actually have users/roles in it.  Maybe rename to `TEST_USERS_WITH_ROLES_PATH`?

selenium_tests/course_info/course_info_base_test_case.py
```diff
@@ -46,52 +58,28 @@ def setUp(self):
...
-
-        if settings.ICOMMONS_REST_API_SKIP_CERT_VERIFICATION:
-            request_args['verify'] = False
-
-        return request_args, url
+            cls.search_page.select_year(year)
+        cls.search_page.submit_search(search_term)
+
+    @classmethod
+    def _load_test_course(cls):
```
rascalking: Is there a reason this needs to be a class method?  All the calls to it in this PR, at least, are done from an instance context.

selenium_tests/course_info/course_info_people_list_tests.py
```diff
@@ -1,31 +1,32 @@
...
         # click on course link to view list of people in course
-        search_page.select_course(cid=course['cid'])
+        self.search_page.select_course(cid=course['cid'])
 
-        people_page = CoursePeoplePageObject(self.driver)
         # assert that an expected enrollment is present
-        self.assertTrue(people_page.is_loaded())
-        self.assertTrue(people_page.is_person_on_page(user['user_id']))
+        self.assertTrue(self.people_page.is_loaded())
+        self.assertTrue(self.people_page.is_person_on_page(user['user_id']))
```
rascalking: Ditto adding the warning about multi-page results.

selenium_tests/course_info/course_info_remove_people_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        # 1. remove ALL roles/enrollments for the test user in this course
+        #    to ensure no incidental data causes conflict when we try to add
+        # 2. add via API the role we want to test removing through the UI
+        self.api.remove_user(course_instance_id, test_user)
+        self.api.add_user(course_instance_id, test_user, role_id)
+
+        self._load_test_course()
+
+        # asserts test user is on people page and to-be-removed user is on page
+        self.assertTrue(self.people_page.is_person_on_page(test_user))
```
rascalking: It's worth adding the "this may fail if there are multiple pages of users for this course" warning from the add users tests.

selenium_tests/course_info/page_objects/course_info_base_page_object.py
```diff
@@ -1,15 +1,7 @@
...
-    This is the base page object class that all Course Info pages can inherit from
-    Locators and Services would be common to all pages on  this tool
-    """
     def __init__(self, driver):
         super(CourseInfoBasePageObject, self).__init__(driver)
-        try:
-            self._driver.switch_to.frame(self._driver.find_element_by_id("tool_content"))
-        except NoSuchElementException:
-            pass
+        self.focus_on_tool_frame()
```
rascalking: This seems like a candidate for moving into selenium common (as `BaseLTIPageObject`, maybe) in the future.

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -68,15 +59,17 @@ def search_and_add_user(self, user_id, role):
         # Enter user to search on
         self.find_element(*Locators.ADD_PEOPLE_SEARCH_TXT).send_keys(user_id)
         # Select role
-        self.select_role_type(role)
+        self.select_role_type(canvas_role)
 
         # Click 'Add to course' course button
         self.find_element(*Locators.ADD_TO_COURSE_BUTTON).click()
+        WebDriverWait(self._driver, 30).until_not(lambda s: s.find_element(
```
rascalking: Just curious...what happens if this times out?

rascalking:
Nothing serious, just some questions and requests for different naming or more comments.

:+1:

selenium_tests/course_info/course_info_base_test_case.py
hktest: Fixed!

selenium_tests/course_info/page_objects/course_info_base_page_object.py
hktest: Let's deal with this in the next round.

selenium_tests/course_info/page_objects/course_people_page_object.py
hktest: Yeah, we ran into the issue where tests were failing because users have not yet been added.  This primarily adds additional "buffer" time (assuming when progress bar disappears, then it's likely user has been added).  If it times out, the next assert should still catch it and throw an error as before -- "assert self.assertTrue(self.people_page.is_person_on_page(test_user))".

selenium_tests/course_info/course_info_remove_people_tests.py
hktest: Good point - added warning/note. Thanks!

selenium_tests/course_info/course_info_people_list_tests.py
hktest: Fixed.

selenium_tests/course_info/page_objects/course_info_base_page_object.py
elliottyates: Re: moving this into selenium common, agreed. On a related note, this is the first time we've cached the POs as TestCase class objects, instead of instantiating them for each new run, so the focus_on_tool_frame isn't as helpful -- since the POs are only instantiated once, in some ways we have to be more careful we're not leaving the tool frame in any tests, or that if we do we refocus at the beginning of the tests. So I think it belongs in selenium common (eventually), but the class-level caching might not work for all test suites...

selenium_tests/course_info/page_objects/course_people_page_object.py
elliottyates: I think this will throw a `TimeoutException` and we won't have to wait until the next assert, in fact. http://selenium-python.readthedocs.org/waits.html

selenium_tests/course_info/course_info_base_test_case.py
hktest: @elliottyates?  (I'll check in with you on this one tomorrow -  there are two usages of _load_test_course on both add/remove, should they be called from the class method instead?  Does it matter?  Newbie-ish (not really sure!)

elliottyates: @hktest and @rascalking: this method doesn't use any instance-specific properties or methods, it merely refers to services provided by class-cached page objects (so it saves each of the instances from making its own version of this routine, if I understand enough about python objects, which I most certainly do not in any complete/authoritative way). @rascalking there's an argument that many of the values passed into this method could in fact be instance attributes, which would clean up some of the method call signatures, but we haven't been doing this with our test cases. Whatever is hitting the sweet spot of efficient, clear, and easy for QA to continue developing with is good by me. So my thought is (a) it shouldn't hold back the PR, so don't wait until tomorrow Hiu Kei if this is stopping you or colleagues, and (b) we can definitely discuss best practices for class/instance methods tomorrow. Thanks both!

rascalking: Oh, absolutely, none of the stuff I brought up should block this getting checked in.

FWIW, I'm used to seeing classmethods being defined/used primarily when they'll need to be called from the class, not from an instance.  ie. I'm used to methods defaulting to instance methods unless there's a reason they can't be (ie. factory methods, singleton classes, etc).

PR #69, opened 2016-02-16
Sprint 41 TLT-2454 Use "'1'" as boolean literal for true

rascalking:
it's the only one that both sqlite and postgres will accept.  probably
the right way to do this in the future is to use RunPython, and work
with models.


elliottyates:
:+1:

PR #70, opened 2016-02-23
Sprint 42 Fix jasmine testing under phantomjs

rascalking:
The issue turned out to be PhantomJS's `String.prototype` missing `startsWith()`.  I've added a new `polyfill.js` under `course_info/tests/libraries`, which will only be pulled in during jasmine testing, let's please confine polyfills needed for testing to that file.

Once that was fixed, I discovered the unit tests were blowing up our `s/Teaching Fellow/TA/` cosmetic change, since it assumed a field existed that didn't when run under unit test.  Fixed that as well.


elliottyates:
@elliottyates

elliottyates:
:+1:

elliottyates:
Great work! This will speed things up and make automated build/testing cycles work again.

PR #72, opened 2016-03-03
Sprint 43: tlt-2449: fixes issues handling course members without profile entries

elliottyates:
- display names are set to the user_id or the search term, as appropriate, where full name is not available from the profile, or blank if duplication would look awkward
- badges fixed to handle missing profile
- ng-switch fixed so that alerts are being displayed properly

plus some refactoring, apologies for the mess.


rascalking:
:+1:

PR #74, opened 2016-03-04
course info add people: fixes issue where some valid email addresses …

elliottyates:
…were not being found

Noticed this while testing 2449. 

- valid, existing email addresses were giving a not found message in the UI due to a limited regex
- this fix switches the regex to match a HUID, and if the match fails assume it's an email (the opposite of the previous behavior)
- the benefit here is that HUID rules are easier to get right than email rules, whereas the downside is that if HUID rules change in the future this code will need to be updated

This is a suggested approach; another would be to find a better javascript regex. An example of a user with a valid email that fails the existing isEmailAddress regex test is `aa011ce0 `.

Also includes some minor refactoring that I apparently could not resist, apologies.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -287,21 +287,25 @@
                 }
             }
         };
-        $scope.isEmailAddress = function(searchTerm) {
-            // TODO - better regex
-            var re = /^\s*\w+@\w+(\.\w+)+\s*$/;
+        $scope.isHUID = function(searchTerm) {
+            var re = /^[A-Za-z0-9]{8}$/;
             return re.test(searchTerm);
         };
         $scope.lookup = function(searchTerm) {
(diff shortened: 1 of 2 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -287,21 +287,25 @@
                 }
             }
         };
-        $scope.isEmailAddress = function(searchTerm) {
-            // TODO - better regex
-            var re = /^\s*\w+@\w+(\.\w+)+\s*$/;
+        $scope.isHUID = function(searchTerm) {
```
rascalking: naming nitpick...that's checking for a valid univ_id, not a huid.

rascalking:
Sounds reasonable to me...validating email addresses is a notoriously tricky thing to get right.  Since @elliottyates is out, I'm going to make the naming change mentioned above and then merge this in.

PR #75, opened 2016-03-08
Task/sapnamysore/tlt 2422/exclude isites filter

sapnamysore:
PR for https://jira.huit.harvard.edu/browse/TLT-2422.

This essentially means we should be filtering out course instances that have been excluded from course sites.[sections in SIS terminology== course instances ]

Jasmine tests to follow in another PR

### DIFF
--- a/course_info/static/course_info/js/controllers/SearchController.js
+++ b/course_info/static/course_info/js/controllers/SearchController.js
@@ -156,6 +156,7 @@
                             $scope.searchInProgress = true;
                         });
                         $scope.enableColumnSorting(false);
+                        //filter the sites flagged to be excluded(exclude_from_isites set to 1)
                         var queryParameters = {
                             exclude_from_isites: 0};
                         if ($scope.queryString.trim() != '') {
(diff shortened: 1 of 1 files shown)


course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -156,7 +156,9 @@
                             $scope.searchInProgress = true;
                         });
                         $scope.enableColumnSorting(false);
-                        var queryParameters = {};
+                        //filter the sites flagged to be excluded(get only ones with exclude_from_isites set to 0)
```
elliottyates: (if possible, let's try to stick to 80-char-width lines, as we've discussed before)

elliottyates:
:+1:

course_info/static/course_info/js/controllers/SearchController.js
sapnamysore: I did think of splitting it in 2 lines, but skipped bcoz it was a comment. Will break it up.

PR #76, opened 2016-03-14
tlt-2511: course info tool course instance list registrar code fix

elliottyates:
Course info search page results table and people page course summary box updates:

- if the registrar_code_display is blank for a course (which sometimes happens in course feeds), falls back on the registrar_code (which we expect to always be present) (AC 1, 2)
- Course Code column no longer includes the course_id field (AC 3)

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -465,7 +465,7 @@
             // and is combination of existing logic in
             // Searchcontroller.courseInstanceToTable and Searchcontroller cell
             // render functions.
-            courseInstance = {};
+            var courseInstance = {};
             if (ci) {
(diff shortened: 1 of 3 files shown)


elliottyates:
(need to code additional AC no. 3. from https://jira.huit.harvard.edu/browse/TLT-2511)

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -465,7 +465,7 @@
             // and is combination of existing logic in
             // Searchcontroller.courseInstanceToTable and Searchcontroller cell
             // render functions.
-            courseInstance = {};
+            var courseInstance = {};
```
rascalking: nice catch.

rascalking:
:+1:

PR #77, opened 2016-03-18
Task/eparker71/tlt 1983/build out course details page

eparker71:
This PR adds a new page called "course details" to the course info tool. Once a user selects a course they get a view of course information. This view is not editable, that is another story. This view has as a field a link to the previously existing people page that shows a list of all the people in the course. 

There is a new html partial, a new controller, and new Jasmine tests.

### DIFF
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -8,7 +8,9 @@
     function DetailsController($scope, $routeParams, courseInstances, $compile,
                                djangoUrl, $http, $q, $log, $uibModal, $sce) {
 
-        dc = this;
+        var dc = this;
+
+        var remove_quotes_regex = new RegExp("^\"|\"$", "g");
 
         dc.handleAjaxError = function (data, status, headers, config, statusText) {
             $log.error('Error attempting to ' + config.method + ' ' + config.url +
@@ -50,12 +52,12 @@
                 .then(dc.handleLookupResults);
         };
 
-        dc.stripQuotes = function(str){
-            // soem fields are coming over with quotes around them and those quotes
-            // are being displayed in the html.
-            // This strips off double quotes from the begining and ending of fields
-            return str ? str.trim().replace(new RegExp("^\"|\"$", "g"), "") : '';
-        };
+        //dc.stripQuotes = function(str){
+        //    // soem fields are coming over with quotes around them and those quotes
+        //    // are being displayed in the html.
+        //    // This strips off double quotes from the begining and ending of fields
+        //    return str ? str.trim().replace(remove_quotes_regex, "") : '';
+        //};
 
         dc.getFormattedCourseInstance = function (ci, members) {
             // This is a helper function that formats the CourseInstance metadata
@@ -64,26 +66,26 @@
             // render functions.
             courseInstance = {};
             if (ci) {
-                courseInstance['title'] = dc.stripQuotes(ci.title);
-                courseInstance['school'] = ci.course ?
-                    ci.course.school_id.toUpperCase() : '';
-                courseInstance['term'] = ci.term ? ci.term.display_name : '';
-                courseInstance['year'] = ci.term ? ci.term.academic_year : '';
-                courseInstance['departments'] = ci.course.departments ? ci.course.departments : [];
-                courseInstance['course_groups'] = ci.course.course_groups ? ci.course.course_groups : [];
-                courseInstance['cid'] = ci.course_instance_id;
-                courseInstance['registrar_code_display'] = ci.course ?
-                ci.course.registrar_code_display +
-                ' (' + ci.course.course_id + ')'.trim() : '';
-                courseInstance['description'] = dc.stripQuotes(ci.description);
-                courseInstance['short_title'] = dc.stripQuotes(ci.short_title);
-                courseInstance['sub_title'] = ci.sub_title ? dc.stripQuotes(ci.sub_title) : '';
-                courseInstance['meeting_time'] = ci.meeting_time ? ci.meeting_time : '';
+                courseInstance['title'] = ci.title;
+                courseInstance['school'] = ci.course.school_id.toUpperCase();
+                courseInstance['term'] = ci.term.display_name;
+                courseInstance['year'] = ci.term.academic_year;
+                courseInstance['departments'] = ci.course.departments;
+                courseInstance['course_groups'] = ci.course.course_groups;
+
+                var registrar_code = ci.course.registrar_code_display ? ci.course.registrar_code_display : ci.course.registrar_code;
+
+                courseInstance['registrar_code_display'] = registrar_code + ' (' + ci.course.course_id + ')'.trim();
+
+                courseInstance['description'] = ci.description;
+                courseInstance['short_title'] = ci.short_title;
+                courseInstance['sub_title'] = ci.sub_title;
+                courseInstance['meeting_time'] = ci.meeting_time;
                 courseInstance['location'] = ci.location;
-                courseInstance['instructors_display'] = ci.instructors_display ? ci.instructors_display : '';
+                courseInstance['instructors_display'] = ci.instructors_display;
                 courseInstance['course_instance_id'] = ci.course_instance_id;
-                courseInstance['notes'] = dc.stripQuotes(ci.notes);
-                courseInstance['conclude_date'] = ci.conclude_date ? ci.conclude_date : '';
+                courseInstance['notes'] = ci.notes;
+                courseInstance['conclude_date'] = ci.conclude_date;
 
                 if (ci.secondary_xlist_instances &&
                     ci.secondary_xlist_instances.length > 0) {
@@ -95,9 +97,9 @@
                     courseInstance['xlist_status'] = 'N/A';
                 }
 
-                courseInstance['sites'] = ci.sites ? ci.sites : [];
+                courseInstance['sites'] = ci.sites;
             }
-            courseInstance['members'] = members.count ? members.count : 0;
+            courseInstance['members'] = members.count;
 
             return courseInstance;
         };
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -1,652 +1,86 @@
-
 // TODO - this is pretty much a copy of the people controller with some mods.
 // TODO - this still needs a lot of clean up, we don't need everything in here
 
-(function() {
+(function () {
     var app = angular.module('CourseInfo');
     app.controller('DetailsController', DetailsController);
 
     function DetailsController($scope, $routeParams, courseInstances, $compile,
-                              djangoUrl, $http, $q, $log, $uibModal) {
-        // set up constants
-        $scope.sortKeyByColumnId = {
-            0: 'name',
-            1: 'user_id',
-            2: 'role__role_name',
-            3: 'source_manual_registrar',
-        };
+                               djangoUrl, $http, $q, $log, $uibModal, $sce) {
 
-        // set up functions we'll be calling later
-        $scope.addUser = function(searchTerm) {
-            $scope.searchInProgress = true;
-            if ($scope.searchResults.length === 0) {
-                $scope.lookup(searchTerm);
-            }
-            else if ($scope.searchResults.length === 1) {
-                $log.error('Add user button pressed while we have a single search result');
-            }
-            else { // $scope.searchResults.length > 1
-                if ($scope.selectedResult.id) {
-                    $scope.addUserToCourse(searchTerm,
-                                           {user_id: $scope.selectedResult.id,
-                                            role_id: $scope.selectedRole.roleId});
-                }
-                else {
-                    $scope.lookup(searchTerm);
-                }
-            }
-        };
-        $scope.addUserToCourse = function(searchTerm, user){
-            var url = djangoUrl.reverse('icommons_rest_api_proxy',
-                                        ['api/course/v2/course_instances/'
-                                         + $scope.courseInstanceId + '/people/']);
-            $scope.clearMessages();
-            $scope.partialFailureData = null;
+        dc = this;
 
-            // called on actual post success, and on error-but-partial-success
-            var handlePostSuccess = function() {
-                $http.get(url, {params: {user_id: user.user_id}})
-                    .success(function(data, status, headers, config, statusText) {
-
-                        // this is a temp fix to change the display text of the role "Teaching Fellow" to "TA"
-                        // a more perm solution is being discussed, but will invlove talking to the schools.
-                        if ( data.results[0].role.role_name == "Teaching Fellow") {
-                            data.results[0].role.role_name = "TA";
-                        }
-
-                        data.results[0].searchTerm = searchTerm;
-                        data.results[0].action = 'added to';
-                        $scope.clearMessages();
-                        // if there was a partial error, specifically if there was
-                        // an error adding the user to the Canvas course (we caught
-                        // a Canvas API error). The user has been added to the
-                        // coursemanager db, but could not be added to Canvas.
-                        // In this case it's possible that the user will be addded
-                        // during the next Canvas sync. We let the user know about
-                        // the partial failure and that it may correct itself.
-                        if($scope.partialFailureData){
-                            data.results[0].partialFailureData = $scope.partialFailureData
-                        }
-
-                        $scope.success = data.results[0];
-                        $scope.dtInstance.reloadData();
-                    })
-                    .error(function(data, status, headers, config, statusText) {
-                        // log it, then display a warning
-                        $scope.handleAjaxError(data, status, headers, config,
-                                statusText);
-                        $scope.clearMessages();
-                        $scope.addPartialFailure = {
-                            searchTerm: searchTerm,
-                            text: 'Add to course seemed to succeed, but ' +
-                                'we received an error trying to retrieve ' +
-                                "the user's course details.",
-                        };
-                    })
-                    .finally(function(){
-                        $scope.clearSearchResults();
-                        $scope.searchInProgress = false;
-                    });
-            };
-
-            $http.post(url, user)
-                .success(handlePostSuccess)
-                .error(function(data, status, headers, config, statusText) {
-                    $scope.handleAjaxError(data, status, headers, config, statusText);
-
-                    if (data.detail &&
-                            (data.detail.indexOf('Canvas API error details') != -1)) {
-                        // partial success, where we enrolled in the coursemanager
-                        // db, but got an error trying to enroll in canvas
-                        $scope.partialFailureData = {
-                            searchTerm: searchTerm,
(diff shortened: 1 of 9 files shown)


course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,107 @@
...
+// TODO - this still needs a lot of clean up, we don't need everything in here
+
+(function () {
+    var app = angular.module('CourseInfo');
+    app.controller('DetailsController', DetailsController);
+
+    function DetailsController($scope, $routeParams, courseInstances, $compile,
+                               djangoUrl, $http, $q, $log, $uibModal, $sce) {
+
+        dc = this;
```
rascalking: Can we scope `dc` to the controller construction function by making this
```js
var dc = this;
```
Otherwise, we're making it a global, which I don't think we should be doing.

eparker71: sure thing

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+            courseInstance = {};
+            if (ci) {
+                courseInstance['title'] = dc.stripQuotes(ci.title);
+                courseInstance['school'] = ci.course ?
+                    ci.course.school_id.toUpperCase() : '';
+                courseInstance['term'] = ci.term ? ci.term.display_name : '';
+                courseInstance['year'] = ci.term ? ci.term.academic_year : '';
+                courseInstance['departments'] = ci.course.departments ? ci.course.departments : [];
+                courseInstance['course_groups'] = ci.course.course_groups ? ci.course.course_groups : [];
+                courseInstance['cid'] = ci.course_instance_id;
```
rascalking: can we please not use `cid`?  it's not clear if you mean "canvas course id" or "course id" or "course instance id".

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+                courseInstance['registrar_code_display'] = ci.course ?
+                ci.course.registrar_code_display +
```
rascalking: I think this needs the same update as in [tlt-2510](https://jira.huit.harvard.edu/browse/TLT-2510)...we want to show registrar_display_code or registrar_code, and never the course id.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+
+            $q.all([coursePromise, membersPromise])
+                .then(dc.handleLookupResults);
+        };
+
+        dc.stripQuotes = function(str){
+            // soem fields are coming over with quotes around them and those quotes
+            // are being displayed in the html.
+            // This strips off double quotes from the begining and ending of fields
+            return str ? str.trim().replace(new RegExp("^\"|\"$", "g"), "") : '';
```
rascalking: super nitpicky, but we can make the regex a controller-level var, and speed this up a wee little bit.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+
+            var membersPromise = $http.get(members_url)
+                .error(dc.handleAjaxError);
+
+            $q.all([coursePromise, membersPromise])
+                .then(dc.handleLookupResults);
+        };
+
+        dc.stripQuotes = function(str){
+            // soem fields are coming over with quotes around them and those quotes
```
rascalking: `s/soem/Some/`

Also, this seems like a feed issue, no?  Have we checked with devops to see if it's something that happens in prod?  I can see a bunch of cases where stripping quotes might not be what we want.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+                courseInstance['course_groups'] = ci.course.course_groups ? ci.course.course_groups : [];
+                courseInstance['cid'] = ci.course_instance_id;
+                courseInstance['registrar_code_display'] = ci.course ?
+                ci.course.registrar_code_display +
+                ' (' + ci.course.course_id + ')'.trim() : '';
+                courseInstance['description'] = dc.stripQuotes(ci.description);
+                courseInstance['short_title'] = dc.stripQuotes(ci.short_title);
+                courseInstance['sub_title'] = ci.sub_title ? dc.stripQuotes(ci.sub_title) : '';
+                courseInstance['meeting_time'] = ci.meeting_time ? ci.meeting_time : '';
+                courseInstance['location'] = ci.location;
```
rascalking: For other fields, we're explicitly checking for a falsey value and using `''` in its place.  Any reason not to stick with that pattern here?  For the record, I'd prefer just putting the values in as is, but either way, I think we should pick one and go with it.

rascalking: Ditto for `course_instance_id` below.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -0,0 +1,237 @@
...
+            var dc = $controller('DetailsController', {$scope: scope});
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify(ci));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify(members));
+            $httpBackend.flush(2);
+            expect(dc.courseInstance['title']).toEqual(ci.title);
+            expect(dc.courseInstance['school']).toEqual(ci.course.school_id.toUpperCase());
+            expect(dc.courseInstance['members']).toEqual(100);
+        });
+
+        it('should work when CourseInstance returns invalid id', function () {
```
rascalking: Maybe this should read `'should log an error when CourseInstance returns an invalid id'`?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -0,0 +1,237 @@
...
+            ci.course_instance_id = 12345;
+            spyOn($log, 'error');
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify(ci));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify(members));
+            $httpBackend.flush(2);
+            expect($log.error).toHaveBeenCalled();
+        });
+    });
+
+    describe('stripQuotes', function () {
```
rascalking: For this method, we don't actually need `ci` or `members`, right?  It takes a string, strips leading/trailing quotes, and returns it.  We do need to instantiate the controller, in order for the method to get defined, but we're not using the data it retrieves.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -0,0 +1,237 @@
...
+            members = {
+                count: 100
+            };
+        });
+
+        it('should strip the quotes from the string', function () {
+            controller = $controller('DetailsController', {$scope: scope});
+            result = dc.stripQuotes(ci.title);
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify(ci));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify(members));
```
rascalking: The `$httpBackend` cleanup has nothing to do with what we're testing...can we move it into an `afterEach` so it's more obvious what's actually involved in the test?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -0,0 +1,237 @@
...
+            // members.count is not defined it should equal 0
+            expect(dc.courseInstance['members']).toEqual(0);
+
+        });
+
+        it('should deal with an empty ci', function () {
+            dc = $controller('DetailsController', {$scope: scope});
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({}));
+            $httpBackend.flush(2);
```
rascalking: Do we log an error if the response is empty?  If so, we should be checking for it here.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
+// TODO - this is pretty much a copy of the people controller with some mods.
```
elliottyates: We're going to clean this up as part of 2376? Or are these comments out-of-date / removable?

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
+{% verbatim %}
+
+<nav>
```
elliottyates: The whole form displays, with empty fields, before the data loads -- it would be good to show something else until the data loads so the user doesn't think it's broken. We could 'bug it' and fix it separately from the PR, or as part of 2376, to keep the stories moving along.

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
...
+        {% endverbatim %}
+        <a href="{% url 'dashboard_account' %}">Admin Console</a>
+        &gt; <a href="#/">Find Course Info</a>
+        {% verbatim %}
+        <small><i class="fa fa-chevron-right"></i></small>
+        Details for {{courseInstance.title}}
+    </h3>
+</nav>
+
+<main class="float=left col=60">
```
elliottyates: This class looks weird to me, is it supposed to have these `=` signs in it and whatnot?

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
...
+                <div class="row">
+                    <div class="col-md-2">
+                        <strong>People: </strong>
+                    </div>
+                    <div class="col-md-10">
+                        <a href="#/people/{{dc.courseInstance.cid}}">
+                            <ng-pluralize count="dc.courseInstance.members"
+                                          when="{'0': 'There are no people in the course.',
+                                'one': 'There is one person in the course.',
+                                'other': 'There are {} people in the course.'}">
```
elliottyates: nice

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </li>
+
+            <li class="list-group-item">
+                <div class="row">
+                    <div class="col-md-2">
+                        <strong>Associated Official and Unofficial Site(s): </strong>
```
elliottyates: ng-pluralize? :)

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: sure, this was mostly copied from the people controller so it might also make sense to update it there as well. I'll put that in.

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
...
+                    </div>
+                    <div class="col-md-10">
+                        <div class="list-group">
+                            <div class="list-group-item list-group-sub-item" ng-repeat="site in dc.courseInstance.sites">
```
elliottyates: At small font sizes in Chrome we get some funky display behavior here... might be worth just 'bugging' it and not troubleshooting it as part of this PR, though.

screen shot 2016-03-18 at 11 00 39 am

course_info/templates/course_info/partials/details.html
```diff
@@ -0,0 +1,211 @@
...
+                                <div class="label label-primary col-md-1" ng-if="site.map_type == 'official'">Official</div>
```
elliottyates: Any reason why we do `ng-switch` for `xlist_status` but `ng-if` here?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+                dc.courseInstance = dc.getFormattedCourseInstance(courseResult.data, membersResult.data)
+            } else {
+                $log.error(' CourseInstance record mismatch for id :'
+                    + dc.courseInstanceId + ',  fetched record for :' + courseResult.data.id);
+            }
+        };
+
+        dc.setCourseInstance = function (id) {
+
+            var course_url = djangoUrl.reverse(
```
rascalking: Most of the time, `courseInstances` should already have the course instance as returned from the list endpoint (it gets populated by the SearchController).  If there's nothing missing from the list serializer that you might need, it would save us an ajax call.

There's also the question of what to do if it *is* missing...eg. if someone just reloaded the lti tool frame.  In the PeopleController, I handled that directly, just like you're doing here.  It's probably worth putting the "fetch if it's missing" logic into the courseInstances directive as a follow-on change once this lands.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+            var members_url = djangoUrl.reverse(
+                'icommons_rest_api_proxy',
+                ['api/course/v2/course_instances/'
+                + id + '/people/']);
+
+            var coursePromise = $http.get(course_url)
+                .error(dc.handleAjaxError);
+
+            var membersPromise = $http.get(members_url)
+                .error(dc.handleAjaxError);
```
elliottyates: Errors are logged fine but we aren't showing anything to the user to indicate a load failed, do we want to do that?

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: yeah, good catch

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: It would be great if the `coursePromise` could initiate display of the data table, and the `membersPromise` could fill in the people info separately, since the `people/` endpoint is slower (the people count section could show a loading indicator until it returned).

course_info/templates/course_info/partials/details.html
rascalking: We can avoid the load most of the time, assuming the course instance list serialization is sufficient here.  But that sounds like a great idea.  Maybe include it in 2376, since it's already the "followon" change?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+                $log.error(' CourseInstance record mismatch for id :'
+                    + dc.courseInstanceId + ',  fetched record for :' + courseResult.data.id);
+            }
+        };
+
+        dc.setCourseInstance = function (id) {
+
+            var course_url = djangoUrl.reverse(
+                'icommons_rest_api_proxy',
+                ['api/course/v2/course_instances/' + id + '/']);
```
elliottyates: Nice attention to detail with adding the `/`!

course_info/templates/course_info/partials/details.html
rascalking: Good catch.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -0,0 +1,110 @@
...
+        dc.handleLookupResults = function(results) {
+            var courseResult = results[0];
+            var membersResult = results[1];
+
+            //check if the right data was obtained before storing it
+            if (courseResult.data.course_instance_id == dc.courseInstanceId) {
+                courseInstances.instances[courseResult.data.course_instance_id] = courseResult.data;
+                dc.courseInstance = dc.getFormattedCourseInstance(courseResult.data, membersResult.data)
+            } else {
+                $log.error(' CourseInstance record mismatch for id :'
```
elliottyates: Ditto on this error -- do we need to display an error message to the user? Pull this out into another feature/bring it into 2376 maybe?

rascalking:
:+1:

course_info/templates/course_info/partials/details.html
elliottyates: So can the course instance list data be referenced in the same scope/passed into the `DetailsController`? That would be great. I can imagine a nice pattern where most of the fields are then immediately available, we show a loading/updating indicator, and we call the details page (and people page) in the background to ensure we have up-to-date information, and finalize the form data once we have those pieces. Every field group could in fact have its own loading indicator predicated upon whether its data point is `null` or `undefined`, and be filled in as need be. Agree that's something to iterate on in 2376 though.

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: As mentioned [here](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/77#discussion_r56673782), I think we'd still want to refetch the details in the background to ensure we have up-to-date information (if they have walked away for a moment and come back). But there's no reason not to show the `courseInstances` list data right away, and load any updates in the background.

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: Good point, if this is a feed issue it should be fixed there. I'm thinking of pulling this from the code and letting the quotes come through. We can look at fixing either the feed or the data if needed.

course_info/templates/course_info/partials/details.html
rascalking: Yup, that's what the `courseInstances` directive is there for...to hand off the cached course instance objects the SearchController already has in memory.

course_info/static/course_info/js/controllers/DetailsController.js
rascalking: Ah, good point.

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: Just had a quick pow pow with @elliottyates . I'm going to pull all the ? true : ''; stuff and let it default to undefined.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: pulling this method

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: pulling this test

course_info/templates/course_info/partials/details.html
eparker71: Yikes, that's ugly. I see it also. It's not immediately clear what's causing it though.

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: there is data missing from the list endpoint, so we do need the separate call at the moment. We could use the data that's there to fill in what we have though.

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: I'd say let's pull it out, if we want to change it, we should also change the people controller so it's bigger than just this story.

PR #78, opened 2016-03-18
Task/josie/tlt 2521 selenium task

Joy603:
Adding additional test for testing registrar code display and bumping up selenium_common version

### DIFF
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -37,15 +37,15 @@ SELENIUM_CONFIG = {
          'cid': '339331',
          'term': 'Spring',
          'title': 'Latin Paleography and Manuscript Culture: Seminar',
-         'school': 'Divinity School',
          'registrar_code_display': '2223',
+         'school': 'Divinity School',
          'type': 'Only courses without sites',
          'year': '2014',
       },
-      'test_course_with_null': {
+      'test_course_with_registrar_code_display_not_populated_in_db': {
          'cid': '353457',
-         'school': 'Harvard College/GSAS',
          'registrar_code_display': 'selenium_test',
+         'school': 'Harvard College/GSAS',
       },
       'test_users': {
          'existing': {
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -44,7 +44,7 @@ SELENIUM_CONFIG = {
       },
       'test_course_with_registrar_code_display_not_populated_in_db': {
          'cid': '353457',
-         'registrar_code_display': 'selenium_test',
+         'registrar_code': 'selenium_test',
          'school': 'Harvard College/GSAS',
       },
       'test_users': {
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -41,6 +41,10 @@ SELENIUM_CONFIG = {
          'type': 'Only courses without sites',
          'year': '2014',
       },
+      'test_course_with_null': {
+         'cid': '353457',
+         'school': 'Harvard College/GSAS',
+      },
       'test_users': {
          'existing': {
             'role_id': '9',
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -38,12 +38,14 @@ SELENIUM_CONFIG = {
          'term': 'Spring',
          'title': 'Latin Paleography and Manuscript Culture: Seminar',
          'school': 'Divinity School',
+         'registrar_code_display': '2223',
          'type': 'Only courses without sites',
          'year': '2014',
       },
       'test_course_with_null': {
          'cid': '353457',
          'school': 'Harvard College/GSAS',
+         'registrar_code_display': 'selenium_test',
       },
       'test_users': {
          'existing': {
--- a/selenium_tests/course_info/course_info_search_tests.py
+++ b/selenium_tests/course_info/course_info_search_tests.py
@@ -20,6 +20,7 @@ class CourseSearchTests(CourseInfoBaseTestCase):
     # If a course does not have a defined registrar_code_display in the db
     # (i.e. if that field is null), then the registrar_code should appear
     # under the 'Course Code' column (TLT-2511)
+    # https://jira.huit.harvard.edu/browse/TLT-2521
 
     def test_registrar_code_when_registrar_code_display_is_null(self):
         """
--- a/selenium_tests/course_info/course_info_search_tests.py
+++ b/selenium_tests/course_info/course_info_search_tests.py
@@ -35,7 +35,7 @@ class CourseSearchTests(CourseInfoBaseTestCase):
             school=course['school'],
             search_term=course['cid'])
 
-        self.assertEqual(self.search_page.get_text(
+        self.assertEqual(self.search_page.get_td_text(
             course['registrar_code']),
             course['registrar_code'],
             "Course Code does not match the expected registrar_code text: "
@@ -53,7 +53,7 @@ class CourseSearchTests(CourseInfoBaseTestCase):
             school=course['school'],
             search_term=course['cid'])
 
-        self.assertEqual(self.search_page.get_text(
+        self.assertEqual(self.search_page.get_td_text(
             course['registrar_code_display']),
             course['registrar_code_display'],
             "Course Code does not match the expected "
--- a/selenium_tests/course_info/course_info_search_tests.py
+++ b/selenium_tests/course_info/course_info_search_tests.py
@@ -15,45 +15,20 @@ class CourseSearchTests(CourseInfoBaseTestCase):
 
         self.assertTrue(self.search_page.is_course_displayed(cid=course['cid']))
 
-    def test_course_code_is_display_when_registrar_code_display_is_null(self):
-        """
-        This test will verify that if the registrar_code_display in the
-        database is null, that the user will see the registrar_code
-        that was entered
-        """
-
-        course = self.test_settings['test_course_with_null']
-
-        self.search_for_course(
-            school=course['school'],
-            search_term=course['cid'])
-
-        self.assertTrue(self.search_page.is_course_code_visible(
-            course['registrar_code_display']))
-
-    def test_course_code_display_when_registrar_code_display_is_not_null(self):
-        """
-        This test will verify that if the registrar_code_display in the
-        database is null, that the user will see the registrar_code
-        that was entered
-        """
-
-        course = self.test_settings['test_course']
-
-        self.search_for_course(
-            school=course['school'],
-            search_term=course['cid'])
-
-        self.assertTrue(self.search_page.is_course_code_visible(
-            course['registrar_code_display']))
+    # If a course has a defined registrar_code_display in the db, then the
+    # registrar_code_display should appear under the 'Course Code' column
+    # If a course does not have a defined registrar_code_display in the db
+    # (i.e. if that field is null), then the registrar_code should appear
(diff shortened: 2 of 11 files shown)


elliottyates:
@elliottyates

selenium_tests/course_info/course_info_search_tests.py
```diff
@@ -14,3 +14,69 @@ def test_course_search(self):
...
+        that was entered
+        """
+
+        course = self.test_settings['test_course_with_null']
+
+        self.search_for_course(
+            school=course['school'],
+            search_term=course['cid'])
+
+        self.assertTrue(self.search_page.is_course_code_visible(
```
elliottyates: I'm not entirely sure these tests do exactly what we're purporting they do; they also seem a bit redundant. There are four tests here, but it seems like we only need two:
- Does the display in the course code column match (exactly) the registrar_code_display for courses that have that field defined? 
- Does the display in the course code column match (exactly) the registrar_code for courses that do not have registrar_code_display defined?

The tests don't seem to be doing that. For one thing, the matches are 'contains', so they're not exact, can we fix that so we're sure we're seeing exactly what we want? It would be great, too, if we could assert that the course code is found in the right td element, instead of just searching for it in the table, in case e.g. the course details column matches it too (although that's unlikely, so just searching for it in the table is probably fine). Finally, I think the four tests can be collapsed into two.

selenium_tests/course_info/course_info_search_tests.py
```diff
@@ -14,3 +14,69 @@ def test_course_search(self):
...
+        course = self.test_settings['test_course']
+
+        self.search_for_course(
+            school=course['school'],
+            search_term=course['cid'])
+
+        self.assertEqual(self.search_page.get_text(
+            course['registrar_code_display']),
+            course['registrar_code_display'],
+            "Registrar code display does not equal")
```
elliottyates: This is a pretty vague error description, can it be more specific? Imagine you were reading test output but didn't know what the test code looked like.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -81,3 +88,35 @@ def select_course(self, cid=None, title=None):
...
+        if element.text > 0:
+            return True
+        return False
+
+    def get_text(self, course_code_text):
+        """
+        :return: The text in the web element in the course code column
+                 else return None
+        """
+        # time.sleep(5)
```
elliottyates: dev cruft?

canvas_account_admin_tools/settings/local.py
```diff
@@ -38,9 +38,15 @@
          'term': 'Spring',
          'title': 'Latin Paleography and Manuscript Culture: Seminar',
          'school': 'Divinity School',
+         'registrar_code_display': '2223',
          'type': 'Only courses without sites',
          'year': '2014',
       },
+      'test_course_with_null': {
```
elliottyates: with null what? can this be a bit more specific?

canvas_account_admin_tools/settings/local.py
```diff
@@ -38,9 +38,15 @@
          'term': 'Spring',
          'title': 'Latin Paleography and Manuscript Culture: Seminar',
          'school': 'Divinity School',
+         'registrar_code_display': '2223',
```
elliottyates: can we keep this sorted alphabetically?

elliottyates:
It would be helpful to get more description in the PR comments about the tests themselves and the feature they're testing.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
Joy603: removed

canvas_account_admin_tools/settings/local.py
Joy603: Done

canvas_account_admin_tools/settings/local.py
Joy603: Changed to test_course_with_registrar_code_display_not_populated_in_db

selenium_tests/course_info/course_info_search_tests.py
Joy603: I updated based on conversation with @elliottyates  and @hktest

selenium_tests/course_info/course_info_search_tests.py
Joy603: Updated

canvas_account_admin_tools/settings/local.py
```diff
@@ -37,10 +37,16 @@
...
          'term': 'Spring',
          'title': 'Latin Paleography and Manuscript Culture: Seminar',
+         'registrar_code_display': '2223',
          'school': 'Divinity School',
          'type': 'Only courses without sites',
          'year': '2014',
       },
+      'test_course_with_registrar_code_display_not_populated_in_db': {
+         'cid': '353457',
+         'registrar_code_display': 'selenium_test',
```
elliottyates: Should this be `registrar_code`?

selenium_tests/course_info/course_info_search_tests.py
```diff
@@ -14,3 +14,47 @@ def test_course_search(self):
             year=course['year'], search_term=course['cid'])
 
         self.assertTrue(self.search_page.is_course_displayed(cid=course['cid']))
+
+    # If a course has a defined registrar_code_display in the db, then the
+    # registrar_code_display should appear under the 'Course Code' column
+    # If a course does not have a defined registrar_code_display in the db
+    # (i.e. if that field is null), then the registrar_code should appear
+    # under the 'Course Code' column (TLT-2511)
+    # https://jira.huit.harvard.edu/browse/TLT-2521
```
elliottyates: Awesome

selenium_tests/course_info/course_info_search_tests.py
```diff
@@ -14,3 +14,47 @@ def test_course_search(self):
...
+
+    def test_registrar_code_when_registrar_code_display_is_null(self):
```
elliottyates: Maybe `test_course_code_when_registrar_code_display_is_null`?

selenium_tests/course_info/course_info_search_tests.py
```diff
@@ -14,3 +14,47 @@ def test_course_search(self):
...
+        """
+        course = self.test_settings[
+            'test_course_with_registrar_code_display_not_populated_in_db']
+
+        self.search_for_course(
+            school=course['school'],
+            search_term=course['cid'])
+
+        self.assertEqual(self.search_page.get_text(
+            course['registrar_code_display']),
```
elliottyates: If registrar_code_display is null in the DB, why are we looking for it, and asserting it isn't null? I think maybe the variable should be named `registrar_code`, right?

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -20,8 +20,14 @@ class Locators(object):
     @classmethod
     def COURSE_LINK_HREF_CSS(cls, cid):
         """ returns a locator for a course detail link in the course table """
-        return By.CSS_SELECTOR, 'a[href="#/people/{}"]'.format(cid)
+        return By.CSS_SELECTOR, 'a[href="#/details/{}"]'.format(cid)
 
+    @classmethod
+    def COURSE_CODE_TEXT(cls, course_code_text):
```
elliottyates: This doesn't match the course code column -- this finds text in any cell in the table. It can be generalized to TD_TEXT(cls, text_to_match_in_td) or something like that. The course people PO has the locator `TD_TEXT_XPATH(cls, search_text)`; this is the same thing. In fact, you could move `TD_TEXT_XPATH()` into the course info base PO so it isn't defined twice.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -81,3 +87,16 @@ def select_course(self, cid=None, title=None):
         # loading the results can take a long time, so explicitly wait longer
         WebDriverWait(self._driver, 30).until(lambda s: s.find_element(
             *CoursePeoplePageLocators.ADD_PEOPLE_BUTTON).is_displayed())
+
+    def get_text(self, course_code_text):
```
elliottyates: This is very vague -- get what text? from where? Also note that it isn't getting the _course code_ text -- it's just matching a TD with the text you send it when you call the method.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
```diff
@@ -81,3 +87,16 @@ def select_course(self, cid=None, title=None):
...
+        """
+        :return: The text in the web element in the course code column
+                 else return None
+        """
+        element = self.find_element(
+            *Locators.COURSE_CODE_TEXT(course_code_text))
+
+        if element.text > 0:
```
elliottyates: I think maybe this should be `if len(element.text) > 0`. And do we want to `strip()` the excess spaces first?

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -25,7 +25,7 @@ def DELETE_USER_ICON (cls, sis_user_id):
     def TD_TEXT_XPATH(cls, search_text):
         """ returns a locator for a table cell element in the people table;
         search_text should be user's name, user_id, etc """
-        return By.XPATH, '//td[contains(text(), "{}")]'.format(search_text)
+        return By.XPATH, '//td[text()="{}")]'.format(search_text)
```
elliottyates: This should match exact text, which is better than `contains()`, but it does depend on your use case -- presumably the other tests that are making use of this PO will still work (that is, their use case is also to search for an exact match).

elliottyates:
Also, in general, this could really be one test with a spreadsheet that had different expectations/test data in it.

elliottyates:
:+1:

selenium_tests/course_info/course_info_search_tests.py
Joy603: Changed

canvas_account_admin_tools/settings/local.py
Joy603: Good catch

selenium_tests/course_info/course_info_search_tests.py
Joy603: changed

selenium_tests/course_info/page_objects/course_info_search_page_object.py
Joy603: Done

selenium_tests/course_info/page_objects/course_people_page_object.py
Joy603: Added this note

selenium_tests/course_info/page_objects/course_info_search_page_object.py
Joy603: Good call.  Done.

selenium_tests/course_info/page_objects/course_info_search_page_object.py
Joy603: Renamed and refactored

Joy603:
@elliottyates :"Also, in general, this could really be one test with a spreadsheet that had different expectations/test data in it." -- We debated over this design initially.  Since an earlier comment, it was suggested that it was sufficient to have 2 tests, unless functionality doesn't change, creating the spreadsheet at this time for testing what gets displayed in 1 column, might seem overkill for a 2 pts story.

elliottyates:
re: overkill -- agreed. It's working now! :)

Joy603:
Resolved merge conflicts and re-tested

PR #80, opened 2016-03-22
Sprint 44: TLT-2376 - edit course details

elliottyates:
Pretty large change from TLT-1983. This PR adds editable fields to the course instance details page in the course_info tool according to TLT-2376 acceptance criteria.

It also:
- streamlines loading (uses existing course instance info if available, pulls extra from API asynchronously, and pulls course members asynchronously)
- provides loading indicators, failure indicators when data is unavailable from backend
- consolidates form fields into angular directives so that changes to field styling can be made in one place

Jasmine test stubs are included, but Jasmine tests can be worked on separately. Thanks!

### DIFF
--- a/course_info/static/course_info/js/app.js
+++ b/course_info/static/course_info/js/app.js
@@ -35,6 +35,7 @@
             .when('/details/:courseInstanceId', {
                 templateUrl: 'partials/details.html',
                 controller: 'DetailsController as dc',
+                //dependencies: ['directives/editable_field']
             })
             .when('/people/:courseInstanceId', {
                 templateUrl: 'partials/people.html',
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -144,7 +144,10 @@
 
             // todo: refactor and collapse, no longer need all these functions since they are single-line
             $http.patch(url, postData)
-                .then($log, dc.handleAjaxError)
+                .then(function finalizeCourseDetailsPatch(response) {
+                    // update Reset button
+                    $.extend(dc.courseInstance, postData);
+                }, dc.handleAjaxError)
                 .finally( function courseDetailsUpdateNoLongerInProgress() {
                     dc.courseDetailsUpdateInProgress = false;
                 });
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -34,8 +34,14 @@
         };
 
         dc.isCourseInstanceEditable = function(courseRegistrarCode) {
+            // TLT-2376: sandbox and ILE courses are editable, and are
+            // identified by their course (registrar) code
             return (courseRegistrarCode.startsWith('ILE-') ||
                     courseRegistrarCode.startsWith('SB-'));
+        };
+
+        dc.isDefined = function(obj) {
+            return (typeof obj === 'undefined');
         };
 
         dc.handleCourseInstanceResponse = function(response) {
@@ -46,8 +52,8 @@
                 // overwrite the members attribute of dc.courseInstance
                 $.extend(dc.courseInstance,
                     dc.getFormattedCourseInstance(response.data));
-                // todo: comment
-                // using 354962 for ILE testing
+                // TLT-2376: only sandbox and ILE courses are currently editable
+                // todo: remove this comment--using 354962 for ILE testing
                 var rc = response.data.course.registrar_code;
                 dc.editable = dc.isCourseInstanceEditable(rc);
                 dc.resetForm();
@@ -156,7 +162,6 @@
                 postData[field] = dc.formDisplayData[field];
             });
 
-            // todo: refactor and collapse, no longer need all these functions since they are single-line
             $http.patch(url, postData)
                 .then(function finalizeCourseDetailsPatch(response) {
                     // update Reset button
@@ -177,6 +182,7 @@
                 editable: '=', // can be < in angular 1.5
                 field: '@',
                 formValue: '=',
+                isLoading: '&',
                 label: '@',
                 modelValue: '=',
             },
@@ -185,10 +191,13 @@
   <div class="form-group"> \
     <label for="input-course-{{field}}" class="col-md-2"> \
       {{label}} \
+      <span ng-show="isLoading()"><i class="fa fa-refresh fa-spin"></i></span> \
     </label> \
     <div class="col-md-10"> \
-      <input type="text" class="form-control" id="input-course-{{field}}" ng-show="editable" ng-model="formValue"/> \
-      <span ng-hide="editable">{{modelValue}}</span> \
+      <div ng-hide="isLoading()"> \
+        <input type="text" class="form-control" id="input-course-{{field}}" ng-show="editable" ng-model="formValue"/> \
+        <span ng-hide="editable">{{modelValue}}</span> \
+      </div> \
     </div> \
   </div> \
 </li> \
@@ -201,6 +210,7 @@
             //templateUrl: 'directives/field_label_wrapper.html'
             scope: {
                 field: '@',
+                isLoading: '&',
                 label: '@',
             },
             transclude: true,
@@ -209,8 +219,10 @@
   <div class="form-group"> \
     <label for="input-course-{{field}}" class="col-md-2"> \
       {{label}} \
+      <span ng-show="isLoading()"><i class="fa fa-refresh fa-spin"></i></span> \
     </label> \
-    <div class="col-md-10" ng-transclude> \
+    <div class="col-md-10"> \
+      <div ng-hide="isLoading()" ng-transclude></div> \
     </div> \
   </div> \
 </li> \
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -8,6 +8,9 @@
                                djangoUrl, $http, $q, $log, $uibModal, $sce) {
 
         var dc = this;
+        // there are two kinds of alerts, global (which appear at the top of the
+        // page/form) and form-level (which appear next to a form label, inside
+        // the form, giving a field-level context to the message)
         dc.alerts = {form: {}, global: {}};
         dc.apiBase = 'api/course/v2/course_instances/';
         dc.apiProxy = 'icommons_rest_api_proxy';
@@ -31,6 +34,13 @@
                     dc.alerts[alertType][alertKey].show);
         };
 
+        dc.isUndefined = function(obj) {
+            // calling angular.isUndefined() directly from the directive
+            // does not seem to work; wrapping it in a function that is watched
+            // on the controller scope works as expected
+            return angular.isUndefined(obj);
+        };
+
         dc.handleAjaxErrorResponse = function(r) {
             dc.handleAjaxError(
                 r.data, r.status, r.headers, r.config, r.statusText);
@@ -48,10 +58,6 @@
                     courseRegistrarCode.startsWith('SB-'));
         };
 
-        dc.isDefined = function(obj) {
-            return (typeof obj === 'undefined');
-        };
-
         dc.handleCourseInstanceResponse = function(response) {
             //check if the right data was obtained before storing it
             if (response.data.course_instance_id == dc.courseInstanceId) {
@@ -65,14 +71,13 @@
                 dc.editable = dc.isCourseInstanceEditable(rc);
                 dc.resetForm();
             } else {
-                $log.error(' CourseInstance record mismatch for id :'
-                    + dc.courseInstanceId + ',  fetched record for :'
+                $log.error('CourseInstance record mismatch for id :'
+                    + dc.courseInstanceId + ', instead received record for '
                     + response.data.id);
             }
         };
 
         dc.handlePeopleResponse = function(response) {
-            if (!dc.hasOwnProperty('courseInstance')) { }
             dc.courseInstance['members'] = response.data.count;
         };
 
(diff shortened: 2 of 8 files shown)


course_info/static/course_info/js/app.js
```diff
@@ -35,6 +35,7 @@
             .when('/details/:courseInstanceId', {
                 templateUrl: 'partials/details.html',
                 controller: 'DetailsController as dc',
+                //dependencies: ['directives/editable_field']
```
rascalking: dev cruft?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -1,70 +1,115 @@
...
-            var courseResult = results[0];
-            var membersResult = results[1];
+        dc.isCourseInstanceEditable = function(courseRegistrarCode) {
+            // TLT-2376: sandbox and ILE courses are editable, and are
+            // identified by their course (registrar) code
+            return (courseRegistrarCode.startsWith('ILE-') ||
+                    courseRegistrarCode.startsWith('SB-'));
+        };
+
+        dc.isDefined = function(obj) {
```
rascalking: angular's got your back here:

https://code.angularjs.org/1.3.9/docs/api/ng/function/angular.isDefined

course_info/static/course_info/js/app.js
elliottyates: dev cruft

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+        };
 
-        dc.setCourseInstance($routeParams.courseInstanceId);
+        dc.resetFormFromUI = function() {
+            dc.resetForm();
+            dc.resetGlobalAlerts();
+            dc.alerts.global.formReset = {show:true};
+        };
 
+        dc.resetGlobalAlerts = function(scrollToAlerts) {
```
rascalking: I know `resetGlobalAlertsAndMaybeScrollToTop()` is awkward, but I wouldn't expect something called `resetGlobalAlerts()` to change the viewport at all.  Can we rename or split it out?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+            // scroll to top of form by default (i.e. if optional scrollToAlerts
+            // argument is not passed in); can disable by setting scrollToAlerts
+            // to false
+            if (scrollToAlerts || scrollToAlerts == null) {
+                scrollTo(0, 0);
+            }
+        };
+
+        dc.submitCourseDetailsForm = function() {
+            dc.courseDetailsUpdateInProgress = true;
```
rascalking: Do we want to set `dc.editable = false` here?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+            var postData = {};
```
rascalking: `patchData`?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+                'title'
+            ];
+            fields.forEach(function(field) {
+                postData[field] = dc.formDisplayData[field];
+            });
+
+            $http.patch(url, postData)
+                .then(function finalizeCourseDetailsPatch() {
+                    // update Reset button
+                    $.extend(dc.courseInstance, postData);
```
rascalking: Angular has its own version of `extend()`...do we want to try to stick to angular-provided util methods over depending more explicitly on jquery?  Or am I just being paranoid?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -1,70 +1,115 @@
...
+                $.extend(dc.courseInstance,
+                    dc.getFormattedCourseInstance(response.data));
+                // TLT-2376: only sandbox and ILE courses are currently editable
+                var rc = response.data.course.registrar_code;
+                dc.editable = dc.isCourseInstanceEditable(rc);
+                dc.resetForm();
             } else {
                 $log.error(' CourseInstance record mismatch for id :'
-                    + dc.courseInstanceId + ',  fetched record for :' + courseResult.data.id);
+                    + dc.courseInstanceId + ',  fetched record for :'
```
rascalking: Can we make this `', instead received record for :'`, so it's more obvious which one we asked for/were expecting, and which one we got?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -1,70 +1,115 @@
...
+                    + response.data.id);
             }
         };
 
-        dc.setCourseInstance = function (id) {
+        dc.handlePeopleResponse = function(response) {
+            if (!dc.hasOwnProperty('courseInstance')) { }
```
rascalking: dev cruft?  we set this in the controller constructor...i would expect this never to try to execute the `{}`, even if something were in it.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -1,70 +1,115 @@
...
+                            details: response.statusText || 'None'};
+                });
 
-            var membersPromise = $http.get(members_url)
-                .error(dc.handleAjaxError);
+            $http.get(members_url)
+                .then(dc.handlePeopleResponse,
+                    function cleanUpFailedCourseMembersGet(response) {
+                        dc.handleAjaxErrorResponse(response);
+                        dc.alerts.form.fetchMembersFailed = {
```
rascalking: Why does the `course_url` error handler clear the global alerts but not this one?

elliottyates: The fetch members failed message is a 'form'-level message -- it appears next to the People label, not at the top of the form. So we don't want to/need to clear global messages, and don't want to scroll (although we could try to scroll to this particular error message, if there's some way of doing it).

screen shot 2016-03-22 at 10 40 44 am

Probably merits more comments at the very least!

course_info/templates/course_info/index.html
```diff
@@ -7,7 +7,7 @@
   <head>
     <meta name="viewport"
           content="width=device-width, initial-scale=1, user-scalable=no">
-    <link href="/static/favicon.ico" rel="icon" type="image/x-icon">
+    <link href="{% static 'favicon.ico' %}" rel="icon" type="image/x-icon">
```
rascalking: nice catch

course_info/templates/course_info/partials/details.html
```diff
@@ -6,208 +6,256 @@ <h3 class="breadcrumbs">
...
-                                    <div class="label label-primary col-md-1" ng-switch-when="unofficial">Unofficial</div>
-                                </div>
-                                <div class="col-md-11"><a href="{{site.course_site_url}}"
-                                                          target="_blank">{{site.course_site_url}}</a>
-                                </div>
-                            </div>
-                        </div>
-                    </div>
+            <hu-editable-input
+                is-loading="dc.isDefined(dc.courseInstance.registrar_code_display)"
```
rascalking: Is there ever a case where `is-loading` isn't just `dc.isDefined(FORM_VALUE)`?  If not, can we just bake that logic into the directive instead of spelling it out every time?

rascalking: Ah, found the case where it isn't.  Blegh...maybe default to `dc.isDefined()` unless overridden?  Kills me to see all that cut-n-paste.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+                'instructors_display',
+                'location',
+                'meeting_time',
+                'notes',
+                'short_title',
+                'sub_title',
+                'title'
+            ];
+            fields.forEach(function(field) {
+                postData[field] = dc.formDisplayData[field];
```
rascalking: Do we want to do this only if the field is dirty?  I guess I'm wondering if the semantics for this page should be "change the things I changed" or "make change everything to match what I see here".

elliottyates: I had the same thought and went with 'change everything' -- given we don't have a confirmation that says "changing the following fields," or show which fields are dirty and have some form instructions letting the user know it won't update everything, I think that user A would expect that if user B changed fields first for that course instance in another browser B, and then user A submits the values in browser A, the course instance should have all of browser A's details. I think this is a big enough issue to make it a separate story/decision, though, as we'd need to communicate what the form was really doing in a way that made sense.

course_info/templates/course_info/partials/details.html
elliottyates: Since it's an isolate scope, you won't have access to `dc.isDefined()` unless it's shared/bound via `&` anyway, but perhaps it's possible to make the default use `angular.isDefined()`... will test it out but I had issues where it didn't bind/update properly when I tried this before -- like I said outside of this PR, I tried lots of ways to refactor... and very much open to cleaning the code up, thanks for the suggestions.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+        return {
+            scope: {
+                editable: '=', // can be < in angular 1.5;
+                               // if not provided, defaults to null/false
+                field: '@',
+                formValue: '=',
+                isLoading: '&',
+                label: '@',
+                maxlength: '@',
+                modelValue: '='
```
rascalking: This is only ever used if the field isn't editable, right?  And we're not trying to pass the value back up out of the isolate scope?  Can it be `'@'`, not `'='`?

rascalking: Can we add a comment explaining the role of `modelValue` vs. `formValue` here?  I'm pretty sure I've figured out what each of them does now, but it took a bit of flipping back and forth between this, the template, and usage of it to get there.

course_info/templates/course_info/partials/details.html
elliottyates: (ah, and `dc.isDefined()` should have been called `dc.isUndefined()` anyway)

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: Yes I think it can be `@`, as we're updating the model in the `finalizeCourseDetailsPatch()`. Nice catch.

course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -57,7 +57,8 @@
                                 text: tc.term_name + ' <span class="caret"></span>',
                             };
                         }));
-                    if (response.data.next !== '') {
+                    if (response.data.next && response.data.next !== '') {
+                        // API returns next=null if there are no more pages
```
rascalking: Hah...that's been annoying me for months.  Thanks for hunting it down, and especially for including the comment explaining it.

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -99,14 +146,129 @@
...
+    </div> \
+  </div> \
+</li> \
+'
+        }
     }
+
+    function fieldLabelWrapperDirective() {
+        return {
+            //templateUrl: 'directives/field_label_wrapper.html'
```
rascalking: dev cruft.

fwiw, i'm guessing the issue was the relative path here is relative to the url of the page, so you'd need to prefix it with `'static/js/'`.  that being said, i'd put it in partials, not in the `js/directives` subdir.

course_info/templates/course_info/partials/details.html
```diff
@@ -6,208 +6,256 @@ <h3 class="breadcrumbs">
...
+                model-value="dc.formDisplayData.registrar_code_display"
+                field="registrar_code_display"
+                label="Course Code"></hu-editable-input>
 
+            <hu-field-label-wrapper
+              field="xlist_status"
+              is-loading="dc.isDefined(dc.courseInstance.xlist_status)"
+              label="Cross listing">
+                <div ng-switch on="dc.courseInstance.xlist_status">
+                  <span ng-switch-when="Primary" class="label label-info">
```
rascalking: Maybe move the `class="label label-info"` up to the switch div so you don't have to repeat it?

course_info/templates/course_info/partials/details.html
elliottyates: FYI I cannot get the default to `angular.isUndefined()` to work -- it requires some combination of accessing the `attrs` and isolate `$scope` via the compile, link, and controller functions for the directive that I cannot quite get right. I'm going to skip this suggestion for the purposes of getting out working code, and we can refactor in parallel with the jasmine tests. I'd be very happy to see these directives improved. One way, as mentioned, is simply to make them use the parent scope, instead of isolate scopes -- then you'd really only have to pass in the `label` and the `field`, and possibly the `editable` and an `isLoading` override -- everything else could be calculated/deduced from the name of the field.

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: Something like that, yes -- looks like we can wrap everything in a `<fieldset>` that will be disable when `dc.courseDetailsUpdateInProgress`.

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: I cannot get this to work either. I think what we really want is `<`, available in the latest version of angular. And I think, given the security implications of being on a very old angular version, we should make moving up to angular 1.5 a refactoring step for this story once the jasmine tests are working.

course_info/static/course_info/js/controllers/DetailsController.js
rascalking: Ooh, I didn't realize `<fieldset>` propagated `disabled` to its children.  Nice!

course_info/static/course_info/js/controllers/DetailsController.js
rascalking: :+1:

rascalking:
:+1:

course_info/static/course_info/js/controllers/DetailsController.js
rascalking: I'm working on pulling down updated versions of all third-party libraries in `course_info/index.html` to icommons-ui now.

PR #83, opened 2016-03-23
tlt-2376: edit course details - fixes course members count 

elliottyates:
(fixes bug TLT-2532)
- the course people tool/page excludes source=xreg course member records; this fix updates the count in the edit course people link to also exclude xreg records


eparker71:
:+1:

PR #84, opened 2016-03-23
Sprint 44 TLT-2376 Revert to using datatables default styling

rascalking:
(depends on https://github.com/Harvard-University-iCommons/django-icommons-ui/pull/79)

So, it turns out the column width issue (sizing them based on the headers, not the data) wasn't due to the styling library, it was because datatables changed the semantics of what it expected for the `deferLoading` option.  That must have been causing some internal error that DT swallowed, which resulted in the bad widths.  Removing it exposed a separate bug, that we were calling `$scope.$apply()` from within our ajax callback, and that might be happening when angular is already in a digest cycle.  Applying the standard "wrap it in `$timeout(..., 0)` instead" fix resolved the issue, and got us back to a sane column width.

We could probably go back and *reapply* the bootstrap styling, but a) I'm tired of messing with this for no real reason, b) a quick straw poll shows people prefer the default stying, and c) this means there should be no user-visible changes from our library upgrades.


elliottyates:
:+1:

PR #85, opened 2016-03-23
Sprint 44 TLT-2376 Add deferLoading back in, fix paths for xlisting

rascalking:
Somehow I missed the fact that with `deferLoading` missing, datatables would immediately load a page of data, before the user picked from the dropdowns and hit search.  Also, I forgot to check the crosslisting app.


elliottyates:
:+1:

PR #86, opened 2016-03-24
Sprint 44 TLT-2520 Add ids to details page for testing

rascalking:

PR #87, opened 2016-03-24
Task/eparker71/tlt 2376/wip on jasmine tests

eparker71:
This PR contains Jasmine tests for TLT-2376

There are a few skipped tests for directives, basically we couldn't figure out how to get the directive and controller data to work together. These have been put in the tech debt item TLT-2539.

### DIFF
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -513,15 +513,5 @@ describe('Unit testing DetailsController', function () {
         it('should show a message if unsuccessful and not update the ' +
             'local course instance data');
     });
-
-    describe('end-to-end tests - form interaction', function() {
-        it('should reset all data when form is reset and show reset ' +
-            'message');
-        it('should call backend on submit');
-        it('should show editable fields and form interaction buttons for ' +
-            'an editable course instance');
-        it('should show non-editable fields for a typical course ' +
-            'instance and no form interaction buttons');
-    });
-
+    
 });
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -451,8 +451,8 @@ describe('Unit testing DetailsController', function () {
         function getCompiledElement(){
             var element = angular.element('<hu-editable-input ' +
                     'editable="dc.editable"' +
-                    'is-loading="true"' +
-                    'form-value="tesValue"' +
+                    'is-loading="dc.isUndefined(dc.courseInstance.term)"' +
+                    'form-value="dc.courseInstance.term"' +
                     'model-value="testModelValue"' +
                     'maxlength="500"' +
                     'field="test-name"' +
@@ -465,7 +465,6 @@ describe('Unit testing DetailsController', function () {
 
         it('should substitute the right stuff for id, name, and label', function(){
             dc = $controller('DetailsController', {$scope: scope});
-            dc.editable = true;
             $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.flush(2);
@@ -479,7 +478,6 @@ describe('Unit testing DetailsController', function () {
             'and the value should be equal to the form\'s copy of the ' +
             'model data', function(){
             dc = $controller('DetailsController', {$scope: scope});
-            //dc.editable = true;
             $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.flush(2);
@@ -491,7 +489,6 @@ describe('Unit testing DetailsController', function () {
             'editable=false or no editable attribute and show the model ' +
             'value, not the form copy of the model value', function(){
             dc = $controller('DetailsController', {$scope: scope});
-            dc.editable = false;
             $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
             $httpBackend.flush(2);
@@ -499,10 +496,11 @@ describe('Unit testing DetailsController', function () {
             expect(liElement.find('input')).not.toBeVisible();
         });
 
-        it('should show only show the label and the loading indicator ' +
+        // skipping this test, can't seem to get it to work.
+        // we can revisit if needed.
+        xit('should only show the label and the loading indicator ' +
             'while loading, and hide the loading indicator when no ' +
             'longer loading', function(){
-
         });
     });
 
@@ -513,5 +511,5 @@ describe('Unit testing DetailsController', function () {
         it('should show a message if unsuccessful and not update the ' +
             'local course instance data');
     });
-    
+
 });
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -295,7 +295,83 @@ describe('Unit testing DetailsController', function () {
     });
 
     describe('submitCourseDetailsForm', function() {
-        it('should call backend with only the editable form fields');
+        var ci;
+        var members;
+        beforeEach(function () {
+            ci = {
+                course_instance_id: $routeParams.courseInstanceId,
+                title: 'Test Title',
+                short_title: 'Test',
+                sub_title: 'Jasime is your friend',
+                description: '<p>hello</p>',
+                sites: [
+                    {
+                        external_id: 'https://x.y.z/888',
+                        site_id: '888',
+                        map_type: 'official'
+                    },
+                    {
+                        external_id: 'https://x.y.z/999',
+                        site_id: '999',
+                        map_type: 'unofficial'
+                    }
+                ],
+                course: {
+                    school_id: 'abc',
+                    registrar_code: 'ILE-2222',
+                    registrar_code_display: '2222',
+                    course_id: '789',
+                    course_groups: [
+                        {
+                            name: 'group one',
+                            id: 1
+                        },
+                        {
+                            name: 'group two',
+                            id: 2
+                        }
+                    ],
+                    departments: [
+                        {
+                            name: 'dept1',
+                            id: 1
+                        }
+                    ]
+                },
+                term: {
+                    display_name: 'Summer 2015',
+                    academic_year: '2015',
+                },
+                primary_xlist_instances: [],
+            };
+
+            members = {
+                count: 100
+            };
+        });
+
+        it('should call backend with only the editable form fields', function(){
+            courseInstances.instances[ci.course_instance_id] = ci;
+            dc = $controller('DetailsController', {$scope: scope});
+            spyOn(dc, 'showNewGlobalAlert');
+            dc.formDisplayData = {
+                'description': 'this is a test',
+                'instructors_display': 'test teacher',
+                'location': 'test room',
+                'meeting_time': 'test oclock',
+                'notes': 'test notes',
+                'short_title': 'test short title',
+                'sub_title': 'test sub title',
+                'title': 'test title'
+            };
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            dc.submitCourseDetailsForm();
+            $httpBackend.expectPATCH(courseInstanceURL).respond(201, JSON.stringify({status: "success"}));
+            $httpBackend.flush(1);
+            expect(dc.showNewGlobalAlert).toHaveBeenCalled();
+        });
         it('should show a message if successful and update the local ' +
             'course instance data');
         it('should show a message if unsuccessful and not update the ' +
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -297,10 +297,21 @@ describe('Unit testing DetailsController', function () {
     describe('submitCourseDetailsForm', function() {
         var ci;
         var members;
+        var expectedPatchData = {
+                "description":"this is a test",
+                "instructors_display":"test teacher",
+                "location":"test room",
+                "meeting_time":"test oclock",
+                "notes":"test notes",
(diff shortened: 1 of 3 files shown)


course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+
+        it('should display the correct value for label', function(){
+            dc = $controller('DetailsController', {$scope: scope});
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            var elem = directiveElem.find('li');
+            expect(elem.find('label').text().trim()).toBe('fieldLabelWrapperDirective test label');
+        });
+        // skipping the following tests, they are tech debt in TLT-2539
```
elliottyates: I think this chunk of `if`s are the ones we agreed to remove completely -- they are effectively E2E (front end) testing that we'll accomplish with a combination of manual + selenium.

elliottyates: Ah, maybe they weren't -- but they don't appear to make much sense in this `describe` block, so I say nuke 'em.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+             inject(function($compile, $rootScope){
+                compile = $compile;
+                scope = $rootScope.$new();
+              });
+              directiveElem = getCompiledElement();
+            scope = directiveElem.isolateScope();
+        });
+
+        function getCompiledElement(){
+            var element = angular.element('<hu-field-label-wrapper ' +
```
elliottyates: FYI these are going to break when #86 goes through, but should be easy to fix up.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            var liElement = directiveElem.find('li');
+            console.log(liElement);
+            expect(liElement.find('input')).not.toBeVisible();
+        });
+
+        // skipping the following test, they are tech debt in TLT-2539
+        xit('should only show the label and the loading indicator ' +
```
elliottyates: Another approach -- rather than `xit`, we could leave these as `it` without an inline function call, which just marks them as 'pending' and doesn't fail the test suite.

elliottyates: cf http://jasmine.github.io/2.0/introduction.html#section-14

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                },
+                term: {
+                    display_name: 'Summer 2015',
+                    academic_year: '2015',
+                },
+                primary_xlist_instances: [],
+            };
+
+            members = {
+                count: 100
```
elliottyates: The actual data is, I think, `members = 100`:
```
dc.courseInstance['members'] = response.data.count;
```
cf https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/develop/course_info/static/course_info/js/controllers/DetailsController.js#L80

elliottyates: my bad, misread the object boundaries, deleting comment about members=100 :)

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+            };
+        });
+
         it('should handle both successful responses even if they arrive ' +
-            'separately');
+            'separately', function(){
+            courseInstances.instances[ci.course_instance_id] = ci;
+            dc = $controller('DetailsController', {$scope: scope});
+            spyOn(dc, 'handleCourseInstanceResponse');
```
elliottyates: We don't use this spy, it can probably be removed. I think it's possibly too late to spy on it anyway -- if we were spying on this successfully and not using `and.callThrough` then line 281 (`expect(dc.courseInstance['title']).toEqual(ci.title)`) shouldn't work, because the real `handleCourseInstanceResponse()` wouldn't be called, no?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify(ci));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify(members));
+            $httpBackend.flush(2);
+            expect(dc.courseInstance['title']).toEqual(ci.title);
```
elliottyates: This isn't actually verifying that all the right info got populated in `dc.courseInstance`. `title` is populated right away because `courseInstances.instances[ci.course_instance_id] = ci` (line 275 in this test, above). In reality, the CI that will be saved in `courseInstances.instances` will have only a subset of the info in `ci`... we should talk about this offline (or I can try to push a patch).

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
         it('should still handle the course instance data even if the ' +
-            'member data fetch fails');
+            'member data fetch fails', function(){
+            courseInstances.instances[ci.course_instance_id] = ci;
+            dc = $controller('DetailsController', {$scope: scope});
+            spyOn(dc, 'handleCourseInstanceResponse');
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify(ci));
+            $httpBackend.expectGET(peopleURL).respond(500);
+            $httpBackend.flush(2);
+            expect(dc.courseInstance['title']).toEqual(ci.title);
```
elliottyates: Ditto here

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                "notes":"test notes",
+                "short_title":"test short title",
+                "sub_title":"test sub title",
+                "title":"test title"
+            };
+
+        beforeEach(function () {
+            ci = {
+                course_instance_id: $routeParams.courseInstanceId,
+                title: 'Test Course Title',
```
elliottyates: Maybe we could define (most of this) in a common function/variable somewhere, instead of declaring it twice in the same suite?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                'title': 'test title'
+            };
+
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            dc.submitCourseDetailsForm();
+            $httpBackend.expectPATCH(courseInstanceURL, expectedPatchData).respond(201, JSON.stringify({status: "success"}));
+            $httpBackend.flush(1);
+            expect(dc.showNewGlobalAlert).toHaveBeenCalled();
```
elliottyates: dc.showNewGlobalAlert _might_ have been called for an error as well. Perhaps we can be more specific, and check that it is called with `'updateSucceeded'`?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                count: 100
+            };
+        });
+
+        it('should call backend with only the editable form fields', function(){
+            courseInstances.instances[ci.course_instance_id] = ci;
+            dc = $controller('DetailsController', {$scope: scope});
+            spyOn(dc, 'showNewGlobalAlert');
+            dc.formDisplayData = {
+                'description': 'this is a test',
```
elliottyates: We could also pull these values from the `ci` so we don't have to update them in two places if they ever changed.

elliottyates: my bad, deleting that update them in two places comment.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                'title': 'test title'
+            };
+
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            dc.submitCourseDetailsForm();
+            $httpBackend.expectPATCH(courseInstanceURL, expectedPatchData).respond(201, JSON.stringify({status: "success"}));
+            $httpBackend.flush(1);
+            expect(dc.showNewGlobalAlert).toHaveBeenCalled();
```
elliottyates: Ditto check for the expected message

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
elliottyates: This test data is defined in multiple places, might be worth refactoring for brevity

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+             inject(function($compile, $rootScope){
+                compile = $compile;
+                scope = $rootScope.$new();
+              });
+              directiveElem = getCompiledElement();
+        });
+
+        function getCompiledElement(){
+            var element = angular.element('<hu-editable-input ' +
+                    'editable="dc.editable"' +
```
elliottyates: This will break with #86 as well I think.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+                    'is-loading="dc.isUndefined(dc.courseInstance.term)"' +
+                    'form-value="dc.courseInstance.term"' +
+                    'model-value="testModelValue"' +
+                    'maxlength="500"' +
+                    'field="test-name"' +
+                'label="test label">' +
+                '</hu-editable-input>');
+            var compiledElement = compile(element)(scope);
+            //scope.$digest();
```
elliottyates: dev cruft?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
+        }
+
+        it('should substitute the right stuff for id, name, and label', function(){
+            dc = $controller('DetailsController', {$scope: scope});
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            var liElement = directiveElem.find('li');
+            expect(liElement.find('label').text().trim()).toBe('test label');
+            expect(liElement.find('input').prop('id')).toBe('input-course-test-name');
```
elliottyates: might as well check `name` too, for completeness

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -212,49 +214,337 @@ describe('Unit testing DetailsController', function () {
...
             'and the value should be equal to the form\'s copy of the ' +
-            'model data');
-        it('should show regular non-editable text if called with ' +
+            'model data', function(){
+            dc = $controller('DetailsController', {$scope: scope});
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+            var liElement = directiveElem.find('li');
+            expect(liElement.find('input').prop('id')).toBe('input-course-test-name');
```
elliottyates: This doesn't appear to be checking the `value` of the input, but just its `id`; also shouldn't we be checking if the `input` is visible and the `span` isn't, to meet the `it()` criteria? That may be hard to do (based on the comments starting on line 489, in the test below this), so we might have to 'tech debt' that part of the test as well.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: removed spyOn

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: yeah, let's talk

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: sure, makes sense

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: gone

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: sure

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: yeah, checking the result of ng-show or ng-hide is not working at least with the toBeVisible() method. I'm not sure why it's not working. We can add a comment here and include that in the tech debt.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: when/if it breaks after code is merged, we can fix it

eparker71:
All the tests work, code had been updated with most of the PR comments. Any refactoring can happen at a later time.

PR #91, opened 2016-03-25
Sprint 44 TLT-2376 Look for the new uib template path

rascalking:
Sorry I missed this.  The angular-ui-bootstrap project changed the path they look for their built-in templates on, and our "don't add `resource_link_id` to the url for uib template urls" logic needed to know about it.


elliottyates:
:+1:

PR #93, opened 2016-03-28
Task/sapnamysore/tlt 2523/selenium test regular course not editable

sapnamysore:
This PR contains selenium tests for verifying that non-ILE courses are not editable.

Separated out the test into it’s own class as this logic was different from the editable tests. And also the test course being loaded in the setUp()method for the other test cases was different(ILE vs non-ILE ).
 Since then, the test has been modified in another branch to separate it out. But I think it can stay separate but happy to convinced otherwise.

### DIFF
--- a/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
+++ b/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
@@ -1,10 +1,7 @@
-from selenium.webdriver.support import expected_conditions as EC
-
 from selenium_tests.course_info.course_info_base_test_case \
     import CourseInfoBaseTestCase
 from selenium_tests.course_info.page_objects.course_info_detail_page_object \
     import  Locators
-
 
 
 class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
@@ -26,14 +23,6 @@ class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
         self.assertEquals(self.detail_page.is_locator_element_present
                           (Locators.SAVE_FORM_BUTTON), False)
 
-        # The tests are validating against an exact match for the class
-        # attribute of non-editable fields
-        # (Editable fields also contain a "ng-binding" but it is always
-        # accompanied by additional suffixes "ng-binding some_value",
-        # which is not an exact match).
-
-        expected_class_name = "ng-binding"
-
         # The span element exists only there if is a value in the field.
         # Limitation:  Each of the fields of the test site is
         # populated with a value. When testing a new course, note that
@@ -52,11 +41,17 @@ class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
         'title'
         ]
 
-        # loops through each of the non-editable fields and looks for the
-        # exact class associated with non-editable fields and also
-        # verifies that the field is not rendered as an input element
+        # The tests are validating against an exact match for the class
+        # attribute of non-editable fields
+        # (Editable fields also contain a "ng-binding" but it is always
+        # accompanied by additional suffixes "ng-binding some_value",
+        # which is not an exact match).
+        expected_class_name = "ng-binding"
+
 
         for element in non_editable_fields:
+            # assert exact class associated with non-editable fields
+            # and also verifies that the field is not rendered as an input element
             self.assertEqual(self.detail_page.get_span_element_class(element),
                              expected_class_name)
             self.assertFalse(
--- a/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
+++ b/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
@@ -1,0 +1,32 @@
+from selenium_tests.course_info.course_info_base_test_case \
+    import CourseInfoBaseTestCase
+from selenium_tests.course_info.page_objects.course_info_detail_page_object \
+    import CourseInfoDetailPageObject, Locators
+
+
+class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
+
+    def setUp(self):
+        super(CourseInfoDetailsNonEditTests, self).setUp()
+        #Load a non-ILE course to validate non editable courses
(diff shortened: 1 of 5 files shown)


selenium_tests/course_info/course_info_details_noneditable_course_tests.py
```diff
@@ -0,0 +1,58 @@
...
+    def setUp(self):
+        super(CourseInfoDetailsNonEditTests, self).setUp()
+        # Load a non-ILE course to validate non editable courses
+        self._load_test_course()
+        self.assertTrue(self.detail_page.is_loaded())
+
+
+    def test_regular_course_not_editable(self):
+        """
+        verify non-SB/ILE courses cannot be edited
```
elliottyates: It would be nice to see the test case (TLT-2523 and test number).

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        # Assert that the Save and Reset button are not present
+        self.assertEquals(self.detail_page.is_locator_element_present
```
elliottyates: Typical page object design is such that the PO's locators and the page's internals are not exposed to the test. Instead, the PO exposes services like `is_course_editable()`, which then use the appropriate Locators internally to return a true or false. The test should not have to know about what the locators are.

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
```diff
@@ -0,0 +1,58 @@
...
+        'term',
+        'title'
+        ]
+
+        # The tests are validating against an exact match for the class
+        # attribute of non-editable fields
+        # (Editable fields also contain a "ng-binding" but it is always
+        # accompanied by additional suffixes "ng-binding some_value",
+        # which is not an exact match).
+        expected_class_name = "ng-binding"
```
elliottyates: Likewise here, this is an implementation detail, and could change -- if it does, you want it encapsulated by the PO, not integrated into the test case. The PO could instead expose a service called `is_field_editable(field_name)`, which would do many of the things here, but it would be encapsulated.

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+        for element in non_editable_fields:
+            # assert exact class associated with non-editable fields
+            # and also verifies that the field is not rendered as an input element
+            self.assertEqual(self.detail_page.get_span_element_class(element),
```
elliottyates: Why are we checking a span class? Each of the span elements has a unique id, i.e. `span-course-{field_name}`, wouldn't it be clearer and sufficient to check for the presence of that element on the page (via a PO service that abstracts the details)?

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -10,6 +12,7 @@ class Locators(object):
     PEOPLE_LINK = (By.ID, "people-link")
     MAIN_TAG = (By.CSS_SELECTOR, "main.course-info-details-page")
     RESET_FORM_BUTTON = (By.ID, "course-details-form-reset")
+    SAVE_FORM_BUTTON = (By.ID, "course-details-form-submit")
```
elliottyates: This is a duplicate of the existing `SUBMIT_FORM_BUTTON` just below.

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -36,12 +47,29 @@ def get_input_field_value(self, field_name):
         input = self.find_element(*Locators.INPUT_BY_FIELD_NAME(field_name))
         return input.get_attribute('value')
 
+    def is_element_displayed_as_input_field(self, field_name):
+        is_displayed = self.find_element(*Locators.INPUT_BY_FIELD_NAME(
```
elliottyates: This will throw a `NoSuchElementException` if not found, instead of returning `False`, right? Do you want to wrap it in a try-except block?

elliottyates:
To stick with the PO pattern, let's try to keep page details encapsulated in the PO. :+1:

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
sapnamysore: Added

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
hktest: As you noted, each element with some value has a unique span element ID, but it's not sufficient enough to just check on the presence of a span element because both editable and non-editable fields have this span element.  Therefore, we condition this way so that the test checks on the presence of an exact class.  "ng-binding" is an exact match that shows up only for non-editable fields.

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
sapnamysore: Will add a specific service checking for  presence of buttons by  moving the locators details into the PO.

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
sapnamysore: Gone :-)

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
sapnamysore: Sure

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
hktest: After further discussion with Sapna, checking the span class isn't the best way to go for this test.  References to span class in both the tests/PO have been removed, and the test has been cleaned up to check on the visibility of the input element instead.

PR #94, opened 2016-03-29
Sprint 44: tlt-2520: course info selenium tests -- fixes broken search tests

elliottyates:
When I resolved conflicts with #78 I did not, in fact, resolve all conflicts! (Short story is that I was testing with some code that was still in my working dir, and missed some of Josie's code that I needed to check in but thought I didn't...) In any case, adding the code back in seems to resolve the issue. Apologies for the confusion.

### DIFF
--- a/selenium_tests/course_info/page_objects/course_people_page_object.py
+++ b/selenium_tests/course_info/page_objects/course_people_page_object.py
@@ -3,7 +3,7 @@ from selenium.webdriver.common.by import By
 from selenium.webdriver.support.wait import WebDriverWait
 
 from selenium_tests.course_info.page_objects.course_info_base_page_object \
-    import Locators as BasePageObjectLocators
+    import Locators as CourseInfoBasePageObjectLocators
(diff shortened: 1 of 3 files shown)


selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -3,6 +3,8 @@
 from selenium.webdriver.support.wait import WebDriverWait
 
 from selenium_tests.course_info.page_objects.course_info_base_page_object \
+    import Locators as BasePageObjectLocators
```
rascalking: I know it's already unwieldy as all hell, but I saw this being used below and went looking for it in selenium common.  Maybe make it `CourseInfoBasePageObjectLocators`?  Or is that too much?

rascalking:
:+1:

selenium_tests/course_info/page_objects/course_people_page_object.py
elliottyates: happy to, for clarity

PR #95, opened 2016-03-29
Sprint 44 TLT-2545 Multiline fields need textareas

rascalking:
Some fields on a course instance may be multiline text, which means we need a `<textarea>`, not an `<input type="text">`.  Add a `<hu-editable-textarea>` directive that mirrors the existing `<hu-editable-input>` one.

bonus: bump the packages used for js unit testing to match those actually in use in the app.

### DIFF
--- a/course_info/static/course_info/js/directives/HuEditableInput.js
+++ b/course_info/static/course_info/js/directives/HuEditableInput.js
@@ -1,0 +1,20 @@
+(function() {
+    var app = angular.module('CourseInfo');
+    app.directive('huEditableInput', huEditableInput);
+
+    function huEditableInput() {
+	return {
+            restrict: 'E',
+	    scope: {
+		editable: '<',  // if not provided, defaults to null/false
+		field: '@',
+		formValue: '=',
+		isLoading: '&',
+		label: '@',
+		maxlength: '@',
+		modelValue: '='
+	    },
+            templateUrl: 'partials/hu-editable-input.html',
+        };
+    }
+})();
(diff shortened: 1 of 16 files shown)


rascalking:
Looks like `npm update` didn't update what I thought it had, and I've got some jasmine tests to fix.  Closing this until that's done.

rascalking:
Reopening with all the directives split out, and working jasmine tests, for real this time.  Had to start using a karma preprocessor plugin (https://github.com/karma-runner/karma-ng-html2js-preprocessor) to avoid complicated `$httpBackend` boilerplate everywhere.  Instead, we get slightly less boilerplate, and a new third-party dependency.

eparker71:
:+1: Looks Great!!!

course_info/static/course_info/js/directives/HuEditableInput.js
```diff
@@ -0,0 +1,20 @@
...
+	    scope: {
+		editable: '<',  // if not provided, defaults to null/false
+		field: '@',
+		formValue: '=',
+		isLoading: '&',
+		label: '@',
+		maxlength: '@',
+		modelValue: '='
+	    },
+            templateUrl: 'partials/hu-editable-input.html',
```
elliottyates: hmmm -- is there some reason this wouldn't have worked for me when I was trying it in the controller file? Seems like the same path and syntax....

rascalking:
Nope... that's why I didn't get the issues you were having.  But I think yours was "directives/...", which didn't have a view that matched it.

PR #98, opened 2016-04-08
Task/eparker71/tlt 2569/crosslisting button visible to ba's only

eparker71:
Added permission check to verify that the current user has permission to see the cross listing button on the account admin dashboard page. If the user does have the correct permission the button will be displayed, otherwise it will not.


sapnamysore:
👍 .  
We should check if  any selenium tests break due to this change.

PR #99, opened 2016-04-08
added selenium test to verify that the cross listing button is displa…

eparker71:
…yed for a user with the correct permissions. In this case, the selenium user has such permissions.

### DIFF
--- a/canvas_account_admin_tools/templates/canvas_account_admin_tools/dashboard_account.html
+++ b/canvas_account_admin_tools/templates/canvas_account_admin_tools/dashboard_account.html
@@ -59,7 +59,7 @@
         </div>
 
       {% if cross_listing_allowed %}
-        <div class="list-group col-md-4">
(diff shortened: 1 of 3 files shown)


canvas_account_admin_tools/templates/canvas_account_admin_tools/dashboard_account.html
```diff
@@ -59,7 +59,7 @@ <h4 class="list-group-item-heading">Find Course Info</h4>
         </div>
 
       {% if cross_listing_allowed %}
-        <div class="list-group col-md-4">
+        <div id="id-cross-listing-div" class="list-group col-md-4">
```
rascalking: Style nitpick...don't think we need to prefix this with `id-`.  Maybe put the id on the button, call it `cross-listing-button`?

rascalking:
:+1:

PR #100, opened 2016-04-12
added tooltips with descriptions for icon messages

eparker71:


rascalking:
👍

elliottyates:
Tooltips do not work for me. Are you sure you don't need to initialize them in the js, as per "Opt-in functionality" (http://getbootstrap.com/javascript/#tooltips-examples)?

rascalking:
Huh...yeah, I ran the branch, and I saw "normal" chrome tooltips, not the bootstrap-looking ones.

elliottyates:
Yeah, so the 'title' is what shows up via normal browser tooltip, and the data-original-title is what is shown in the bootstrap-styled tooltips.

eparker71:
yeah, they need to be initialized. I thought they were from the base already, but it was just the chrome ones.

PR #101, opened 2016-04-13
Task/sapnamysore/tlt 2567/xlist ordering

sapnamysore:
- Modified the ordering params based on https://jira.huit.harvard.edu/browse/TLT-2567 .
Depends on [this](https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/62) PR for backend.
- Added a spinning indicator while the  list gets loaded

### DIFF
--- a/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
+++ b/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
@@ -4,8 +4,8 @@
      */
     var app = angular.module('CrossListCourses');
     app.controller('ListController',
-        ['$scope', '$http', '$timeout', '$document', '$window', '$compile','djangoUrl',
-        function($scope, $http, $timeout, $document, $window, $compile, $djangoUrl){
+        ['$scope', '$http', '$timeout', '$document', '$window', '$compile','djangoUrl','$log',
+        function($scope, $http, $timeout, $document, $window, $compile, $djangoUrl, $log){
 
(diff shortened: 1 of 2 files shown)


eparker71:
:+1:

cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
```diff
@@ -60,13 +60,15 @@
                         next: '',
                         previous: '',
                     },
+                    processing: '<img src="' + window.globals.STATIC_URL + 'images/ajax-loader-large.gif" class="loading"/> Loading...',
```
elliottyates: Any reason we went with this instead of the fontawesome one we're using elsewhere? https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/37f7596e569967b265727d583fb2b0491bab4f4f/course_info/templates/course_info/partials/details.html#L11

elliottyates: (It would avoid the extra static URL thing -- but perhaps it doesn't work well with datatables?)

sapnamysore: Went by what the PO pointed at (in the Bulk Site Creator), didn't recollect that we had the fontawesome one.  Happy to switch, but she wanted the large spinner as opposed to the smaller one. I can check in with Chrissy and @Vittorio2015 and switch. Thanks for pointing it out.

PR #102, opened 2016-04-14
add people confirmation message

eparker71:
Added a confirmation message to be displayed before people added to the course.

All of the original add people code is intact even though it looks like it changed. I just wrapped it in a modal dialog. I know that wasn't the original design but it turned out to be a cleaner implementation to do it this way. If it becomes an issue we can change it. 

removed table-hover from both tables to address trello task "Results table should not have mouseover capability"

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -127,42 +127,68 @@
                 .then(handlePostSuccess, handlePostError);
         };
         $scope.addPeopleToCourse = function(searchTerms) {
+
             /* looks up HUIDs, XIDs, and/or email addresses from searchTerms
              and attempts to add people to the course who do not already have an
              enrollment.
              */
-            var membersByUserId = {};
-            $scope.clearMessages();
-            $scope.operationInProgress = true;
-            var searchTermList = $scope.getSearchTermList(searchTerms);
-            $scope.tracking.total = searchTermList.length;
-            $scope.updateProgressBar('Looking up ' + $scope.tracking.total
-                + ' people');
-            var memberPromise = $scope.lookupCourseMembers()
-                .then(function updateCourseMembers(memberResponse) {
-                    membersByUserId = $scope.getMembersByUserId(
-                        memberResponse.data.results);
-                    return memberResponse;
-                }, function courseMemberLookupFailed(memberResponse) {
-                    $scope.handleAjaxErrorResponse(memberResponse);
-                    // todo: how do we make this break the add chain?
-                    return $q.reject('Course member lookup failed');
-                });
-            var peoplePromises = $scope.lookupPeople(searchTermList);
-            var addNewMemberPromises = [];
-            peoplePromises.forEach(function setupAddPersonPromiseChain(personPromise) {
-                addNewMemberPromises.push(
-                    $q.all([memberPromise, personPromise])
-                        .then(function addFetchedPerson(responses) {
-                            var personResponse = responses[1];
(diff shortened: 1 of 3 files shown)


elliottyates:
@eparker71 I apologize, I didn't realize you had put in this PR -- can you please rebase this and resolve the conflicts?

eparker71:
no problem

sapnamysore:
👍

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -120,49 +120,76 @@
...
-                    membersByUserId = $scope.getMembersByUserId(
-                        memberResponse.data.results);
-                    return memberResponse;
-                }, function courseMemberLookupFailed(memberResponse) {
-                    $scope.handleAjaxErrorResponse(memberResponse);
-                    $scope.messages.warnings.push({
-                        type: 'courseMemberLookupFailed'
+
+            // split the list to get the number of people the user is trying to add
+            $scope.numPeople = searchTerms.split(',').length;
```
elliottyates: `$scope.getSearchTermList(searchTerms)` encapsulates this (see line 155, below), and includes splitting by newlines as well. We can move that up to here instead.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -120,49 +120,76 @@
...
+            var modalInstance = $uibModal.open({
+                animation: true,
+                templateUrl: 'partials/add-people-to-course-confirmation.html',
+                controller: function ($scope, $uibModalInstance, numPeople, selectedRoleName) {
+                    $scope.numPeople = numPeople;
+                    $scope.selectedRoleName = selectedRoleName;
+                },
+                resolve: {
+                    numPeople: function () {
+                        return $scope.numPeople;
```
elliottyates: I don't think these need to be global scope variables/members, they can just be local and not pollute the `PeopleController` scope.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -120,49 +120,76 @@
...
+                    },
+                    selectedRoleName: function () {
+                        return $scope.selectedRoleName;
+                    }
+                }
+            });
+
+            modalInstance.result.then(function modalSuccess() {
```
elliottyates: To avoid further indentation/nesting, we can just call a modified version of the original `AddPeopleToCourse()` and put all the modal stuff in its own scope-level function.

course_info/templates/course_info/partials/people.html
```diff
@@ -261,7 +260,7 @@ <h3 class="breadcrumbs">
 
   <!-- begin course membership interface -->
   <table datatable="" dt-options="dtOptions" dt-columns="dtColumns"
-         dt-instance="dtInstance" class="table table-striped table-hover">
+         dt-instance="dtInstance" class="table table-striped">
```
elliottyates: I think the intention was that only the warnings/failures table should not have hover styling; the course membership table should continue to have hover styling, as you can actually perform actions on those (i.e. delete).

elliottyates:
Looks great and matches the behavior of the confirmation of remove people, as well, come to think of it.

elliottyates:
FYI here's an example of the PR suggestions, feel free to check it over/merge into this branch before merging this branch into the main 2119 collaboration branch. Nice work! https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/103

eparker71:
I updated the code per PR suggestions and merged in the #103 branch and made sure to rebase off the original.

PR #104, opened 2016-04-15
Task/hchow/tlt 2574 multiple add users

hktest:
Adding selenium tests for multiple add functionality in Course Info Tool.  

Focusing on the multi-add only.  There are a few broken existing single-add due to locator changes - fixed those as well.


selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -67,11 +77,38 @@ def select_role_type(self, canvas_role):
...
+    def add_is_successful_by_id(self, user_id):
+        # Verify user ID is displayed
+        try:
+            self.find_element(*CourseInfoBasePageObjectLocators.TD_TEXT_XPATH(
+                        user_id)).is_displayed()
+        except NoSuchElementException:
+            return False
+        return True
+
+    def multiple_add_partial_failure(self):
```
elliottyates: I think we could have the single & multiple in the same service, and send in the number of successes we expect.

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -67,11 +77,38 @@ def select_role_type(self, canvas_role):
...
+            self.find_element(*Locators.SINGLE_ADD_ALERT_TEXT)
+        except NoSuchElementException:
+            return False
+        return True
+
+
+    def add_is_successful_by_id(self, user_id):
+        # Verify user ID is displayed
+        try:
+            self.find_element(*CourseInfoBasePageObjectLocators.TD_TEXT_XPATH(
```
elliottyates: Successful adds are not displayed in the results table, if that's what we're attempting to match here. They _are_ displayed in the course members list, but may not be on the current page, depending on who's in the course and what the page size and ordering is, etc.

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -51,3 +52,94 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+        """
+        self._load_test_course()
+        self.detail_page.go_to_people_page()
+        self.assertTrue(self.people_page.is_loaded())
+
+
+    def test_multi_user_add_unsuccessful(self):
+        """
+        TLT-2574
+        This test verifies that multiple_user_add (for invalid ID) is
```
elliottyates: Any reason this looks like a variable name? (`multiple_user_add`) -- is it meant to be descriptive, or refer to an actual piece of code?

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -51,3 +52,94 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+        People's page
+        """
+        self._load_test_course()
+        self.detail_page.go_to_people_page()
+        self.assertTrue(self.people_page.is_loaded())
+
+
+    def test_multi_user_add_unsuccessful(self):
+        """
+        TLT-2574
```
elliottyates: These are helpful! If we can refer to actual test case numbers too, that would be great.

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -51,3 +52,94 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+    def _load_find_info_tool(self):
+        """
+        Common code: This method loads up the find_info_tool and goes to the
+        People's page
+        """
+        self._load_test_course()
+        self.detail_page.go_to_people_page()
+        self.assertTrue(self.people_page.is_loaded())
+
+
```
elliottyates: nit pick -- typically only one blank line between methods. (PEP-8 -- PyCharm should show you this stuff as yellow warnings)

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -51,3 +52,94 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
         self.api.remove_user(self.test_settings['test_course']['cid'],
                              test_user, role_id)
 
+
+class MultipleAddPeopleTests(CourseInfoBaseTestCase):
+
+    def _load_find_info_tool(self):
+        """
+        Common code: This method loads up the find_info_tool and goes to the
```
elliottyates: I'm unfamiliar with this terminology, what is the find_info_tool?

canvas_account_admin_tools/settings/local.py
```diff
@@ -57,6 +58,13 @@
          'type': 'All courses',
          'year': '2016',
       },
+      # TLT-2574 - Course Info People Add
+      'test_data_for_multiple_users_add': {
+         'canvas_role': 'Guest',
+         'unsuccessful_add': '12345678, 23456789',
```
elliottyates: This is probably meant to be a list?

hktest:
Note: Commit  ac96de9 fixed the failures in existing single add user tests (due to locator changes)

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Ah, good point.  Not referring to code, thought it was descriptive.  Re-worded it in all the comments that had this reference.

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Updated.  Generally, I'll stick to referring to AC as it is more stable.  Will refer to specific TC when applicable (talk to Robert - he will be expanding on a few), but keep in mind that TC numbering is generally more fluid than AC (so the numbering could be changed when TC's are added/removed/re-ordered).   Fixed for these tests - thanks!

selenium_tests/course_info/course_info_add_people_tests.py
hktest: that's my bad, and a good nitpick   (I don't - personally - agree with pep-8 on one blank line because that looks crowded to me ;)  but I'll stick to that standard - thanks!

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Hmm, point taken.  Also not really accurate in terms of description.  Fixed!

canvas_account_admin_tools/settings/local.py
hktest: Nope, this one is meant to be a string (and is working as intended).  This is for the unsuccessful add, so we can just pass in a string of however many ids we want to pass in to the add user input field.  Unlike the other tests, there is no need to put it into a list and then join it, because there is no  clean up needed for this test (via rest api call, which removes it one at a time, where the list is needed).  So I want to pass this in as a string.  Confusing much, perhaps in the context of the other two data points?  I added comments already within the test itself, but I can add in a comment here if that's helpful...

hktest: And, I'm going to re-order alphabetically as well.  :)

selenium_tests/course_info/page_objects/course_people_page_object.py
hktest: Indeed, this is checking against the course members list that appears on the page (not the results table).  It's a little bit of a more precise check than just the alert text, but you're right regarding the pagination.   There were a bunch of TODOs carryover in the code for the single create that "the test methods weren't actually checking on actual ID", so I thought I get my practice in on having a method that checks for ID (as a reference for future if needed).  

We could go either way really.  Considering most of tests does the cleanup, it may not be an issue?  I had already added a note in the multi_user_add_successful test:   re: "''' Limitation:  This may no longer be a valid test if there are a lot of ID in the test course and the user being added appears on second page due to pagination.  If that is the case, we could go back to checking on alert text'''" 

Let me know if you have particularly strong feelings about using alert text instead (due to potential flakiness factor, if that's the biggest concern) and I could modify it.  Thanks!

selenium_tests/course_info/page_objects/course_people_page_object.py
hktest: Do you mean combining the successful page object methods: 
 single_add_was_successful() and add_is_successful_by_id()?

One checks by id and one checks by alert text.  See related comment below.  I can fix it once we decide which one we want to check success by.

selenium_tests/course_info/page_objects/course_people_page_object.py
elliottyates: No problem -- ok in that case, let's just make sure that any tests which are asserting this is `True` at the end also assert it is `False` at the beginning, so we aren't mistakenly assuming the add worked when in fact the ID was already in the course for some reason.

selenium_tests/course_info/page_objects/course_people_page_object.py
elliottyates: Sorry, I meant `multiple_add_partial_failure()` (actually, that method name doesn't appear to be accurate anyway?) and `single_add_was_successful()`. They are both checking the message, but differ slightly -- it might be clearer if the service was something like `num_of_people_added_successfully(x)`, and in that method if `x==1` we check for the single add success message, if `x>1` we check for the multiple add success message.

canvas_account_admin_tools/settings/local.py
elliottyates: No no! I misread it. :) I though it read `'x', 'y'` instead of `'x, y'`. My bad, thanks for the clarification!

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+
+        # Join the id_list so we can pass in test users as a string; not list
+        element = ', '.join(id_list)
+        # Search and add valid ID to the course
+        self.people_page.search_and_add_user(element, self.test_data['canvas_role'])
+
+        ''' Limitation:  This may no longer be a valid test if there are a lot
+        of ID in the test course and the user being added appears
+        on second page due to pagination.  If that is the case, we could go
+        back to checking on alert text'''
```
elliottyates: True -- some other approaches:
- change the page size to 100 before adding (doesn't eliminate the possiblity entirely, though)
- 'guard' against this by checking the page size and the number of people in the course before attempting the add (so if `num_of_people >  page_size - people_to_add` then we fail the test with an explanatory error, to make it easier to diagnose and fix)

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+            self.api.remove_user(self.test_settings['test_course']['cid'],
+                                 user_id)
+
+        # Load up the test course and go to People Page
+        self._load_find_info_tool()
+
+        # Join the id_list so we can pass in test users as a string; not list
+        element = ', '.join(id_list)
+        # Search and add valid ID to the course
+        self.people_page.search_and_add_user(element, self.test_data['canvas_role'])
```
elliottyates: We should probably rename `search_and_add_user()` to `search_and_add_users()` (i.e. pluralize it)

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+        # Remove test user if they are already in course.
+        for user_id in id_list:
+            self.api.remove_user(self.test_settings['test_course']['cid'],
+                                 user_id)
+
+        # Load up the test course and go to People Page
+        self._load_find_info_tool()
+
+        # Join the id_list so we can pass in test users as a string; not list
+        element = ', '.join(id_list)
```
elliottyates: `element` is a bit confusing here, maybe something like `user_id_input_string`?

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+        self.people_page.search_and_add_user(element, self.test_data['canvas_role'])
+
+        ''' Limitation:  This may no longer be a valid test if there are a lot
+        of ID in the test course and the user being added appears
+        on second page due to pagination.  If that is the case, we could go
+        back to checking on alert text'''
+
+        # Verify successful add (if ID appears in the list of users)
+        for user_id in id_list:
+            self.people_page.add_is_successful_by_id(user_id)
```
elliottyates: `self.assertTrue()` ?

elliottyates:
Can we remove `deprecated_admin_console_add_multi_users.xlsx` completely?

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -67,11 +77,38 @@ def select_role_type(self, canvas_role):
...
-        # todo: this does not check _which_ add was successful...
+        # todo: we can refactor this; see "add_is_successful_by_id" method
+        try:
+            self.find_element(*Locators.SINGLE_ADD_ALERT_TEXT)
+        except NoSuchElementException:
+            return False
+        return True
+
+
+    def add_is_successful_by_id(self, user_id):
```
elliottyates: Isn't this really just the same thing as `is_person_on_page()`?

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -9,15 +9,25 @@
...
-    ALERT_SUCCESS_DELETE_PERSON = (By.XPATH, '//p[contains(.,"was just removed")]')
+    ADD_PEOPLE_SEARCH_TXT = (By.ID, 'emailHUID')
+    ADD_TO_COURSE_BUTTON = (By.ID, 'add-user-btn-id')
+    ALERT_SUCCESS_ALERT_BOX = (By.ID, 'alert-success')
+    ALERT_UNSUCCESSFUL_ADD_PERSON = (By.XPATH, '//div[contains(.,"could not '
+                                               'be added")]')
+    ALERT_SUCCESS_DELETE_PERSON = (By.XPATH, '//p[contains(.,"just removed")]')
     DELETE_USER_CONFIRM = (By.XPATH, '//button[contains(.,"Yes, Remove User")]')
+
+    MULTI_USER_PARTIAL_FAILURE_TEXT = (By.XPATH, '//div[contains(.,"person '
```
elliottyates: This would only match if 1 person was not added -- seems like a flexible approach would be to collapse an equivalent to [the logic in the frontend](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/task/elliottyates/tlt-2119/add_people_allow_multiple/course_info/templates/course_info/partials/people.html#L153) into services on this PO called something like `people_added(n)` and `people_not_added(n)`, which will encapsulate checking for the right messages given the number of people.

selenium_tests/course_info/course_info_base_test_case.py
```diff
@@ -51,6 +51,9 @@ def setUpClass(cls):
         cls.people_page = CoursePeoplePageObject(cls.driver)
         cls.search_page = CourseSearchPageObject(cls.driver)
 
+        cls.test_data = settings.SELENIUM_CONFIG[
+            'course_info_tool']['test_data_for_multiple_users_add']
```
elliottyates: Does it make sense for the entire course info base test case to use `test_data_for_multiple_users_add` as its test data? Seems like that's very specific.

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+
+        #  Load up test course and go to People page.
+        self._load_find_info_tool()
+
+        # Join the id_list so we can pass in test users as a string; not list
+        element = ', '.join(id_list)
+        self.people_page.search_and_add_user(element,
+                                             self.test_data['canvas_role'])
+        # Verify that user is not added through alert text
+        self.assertTrue(self.people_page.multiple_add_partial_failure())
```
elliottyates: I think `multiple_add_partial_failure()` is a difficult name to parse. Recall that the idea around POs is that they present services to test cases. What does a service called `multiple_add_partial_failure()` do, exactly? It's not clear to me whether it does something, or checks something, or navigates, etc. Further, it's such a specific thing -- it's not really a reusable component / service, but rather it is in itself a unique test case -- that is, the service returns true if and only if one user was added successfully and one user was not, so it's not very extensible. Perhaps it would be best to refactor that into a more reusable service (see suggestions for that in other comments).

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+
+
+    def test_multiple_add_partial_failure(self):
+        """
+        TLT-2574:
+        This test verifies that multiple_user_add is partially successful.
+        Some added.  Others aren't. Cleanup of data includedincluded.
+        """
+        # test data as a list, since rest api removes user one at a time
+        id_list = self.test_data['partial_success_add']
```
elliottyates: This method and `test_multi_user_add_successful()` have a lot of common code. Perhaps they can be collapsed in some way to reduce duplication?

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -44,10 +45,100 @@ def test_add_person(self, test_case_id, test_user, canvas_role, role_id):
...
+        self.people_page.search_and_add_user(self.test_data['unsuccessful_add'],
+                                             self.test_data['canvas_role'])
+        # Verify that unsuccessful message shows up
+        self.assertTrue(self.people_page.add_was_unsuccessful())
+
+    def test_multi_user_add_successful(self):
+        """
+        TLT-2574:
+        This test verifies that multiple_user_add (for valid ID) are successful.
+        Cleanup of data required and included via rest API calls.
```
elliottyates: This kind of information is helpful.

elliottyates:
A suggestion to refactor the PO service into a single, reusable function: https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/107

selenium_tests/course_info/course_info_add_people_tests.py
hktest: To be discussed:  Is collapsing the duplication here still relevant if I'm going to "refactor the PO service into a single, reusable function"? (or is this one and same/related comment?)

selenium_tests/course_info/course_info_base_test_case.py
hktest: Will look into moving it into the test level (related to tests), rather than in the project's base_test_case.

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Fixed, and nice catch - thanks!

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Fixed.

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Fixed.

selenium_tests/course_info/page_objects/course_people_page_object.py
hktest: Yes and fixed!  Removed duplicate method.

hktest:
@elliottyates - code wise, i'm done with making any more code changes to fix any broken tests (fixed data, broken tests when running regression_suite).  I'm able to run regression_suite.py successfully locally.  when i run this on jenkins, it all passed except for the occasional pesky stale element exception --- all seems to be in the addPeople tests, failing at the progress bar search.  I'm still poking around that a little bit to see if I can make sure progress, but other than that, it's ready for PR.

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -2,52 +2,52 @@
...
-        # ensure person is not in course before attempting to add using API;
-        # remove ALL roles/enrollments for the test user in this course
+        # ensure people are not in course before attempting to add using API;
+        # remove ALL roles/enrollments for the test users in this course
         # to ensure no incidental data causes conflict when we try to add
-        self.api.remove_user(self.test_settings['test_course']['cid'],
-                             test_user)
 
-        self._load_test_course()
+        test_course_key = 'test_course_SB_ILE'
```
elliottyates: Now that you've [changed the test_course](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/104/files#diff-161fca96bd4a31782cda875c287a0601R37) do we just want to be using that, instead of `test_course_SB_ILE`?

selenium_tests/course_info/course_info_add_people_tests.py
```diff
@@ -2,52 +2,52 @@
...
 
         self.detail_page.go_to_people_page()
 
         # search for a user and add user to course
         self.assertTrue(self.people_page.is_loaded())
 
-        # assert that the success message is not already loaded on the page
-        self.assertFalse(self.people_page.add_was_successful())
+        # assert that the results message is not already loaded on the page
+        self.assertFalse(self.people_page.people_added(int(successes), int(failures)))
```
elliottyates: We _might_ be able to remove the `int()` wrappers if we updated the spreadsheet, but even the Excel 'number' style might come through as a `float`. Not sure, so not a big deal -- it works this way and isn't too unwieldy.

selenium_tests/course_info/course_info_remove_people_tests.py
```diff
@@ -13,7 +13,8 @@ class RemovePeopleTests(CourseInfoBaseTestCase):
     def test_remove_person(self, test_case_id, test_user, canvas_role, role_id):
         """ Removes a user from course using the Admin Console """
 
-        course_instance_id = self.test_settings['test_course']['cid']
+        test_course_key = 'test_course_SB_ILE'
```
elliottyates: See https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/104/files#r60489715

selenium_tests/course_info/page_objects/course_people_page_object.py
```diff
@@ -40,41 +46,65 @@ def is_person_removed_from_list(self, lookup_text):
         user id
         """
         try:
-            WebDriverWait(self._driver, 30).until_not(lambda s: s.find_element(
+            WebDriverWait(self._driver, 60).until_not(lambda s: s.find_element(
```
elliottyates: Is this the part that was failing? I'm not sure what you removed to fix the stale element exceptions.

elliottyates:
👍 keep in mind this won't work against the _current_ state of the develop branch of canvas_account_admin_tools -- but [the PR is in](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/112) for the updates required to make this work!

selenium_tests/course_info/course_info_add_people_tests.py
hktest: Fixed!

selenium_tests/course_info/course_info_remove_people_tests.py
hktest: Fixed here too.  Scanned through all the test files to see if I modified it anywhere else - seems these are the only two where test course had been modified. Thanks!

selenium_tests/course_info/page_objects/course_people_page_object.py
hktest: this extension of time is to give test a little more time to locate element which sometimes fails (why is why there is a webdriver wait  in there in the first place).  The stale element exception is in the removal of the progress check bar lines 62-63 here:  https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/104/files#r60489759

selenium_tests/course_info/course_info_add_people_tests.py
hktest: I vote for keeping the int() wrappers in there.  I did a quick test on converting the cell to "numbers", but the test fails out on that line.  It's not too unwieldy, so let's keep as is for now.  Thanks!

hktest:
Pulled the latest from develop and re-ran regression_suite.py (off Jenkins).  All passed - merging in.

PR #106, opened 2016-04-15
changed the "Add People" button from an anchor tag to a button (it wa…

eparker71:
…s a button anyway) so it can be tabbed to. the page was not allowing tab to the anchor.


elliottyates:
👍

PR #108, opened 2016-04-18
041816 fix up stale element failure in remove tests

hktest:
canvas_account_course_admin project has not had a successful run in Jenkins since late March.

There appears to be a persistent failure in the "course_info_remove_people_tests" -  it's failing due to "StaleElementReferenceException: Message: Element not found in the cache - perhaps the page has changed since it was looked up"

Error is relatively persistent;  the test user at which it failed on is not at all consistent.  This issue is  reproducible on Jenkins and locally.  Adding the code fixes this issue.   I have not been able to reproduce the stale_element error with this change locally and when I run it on Jenkins anymore.

### DIFF
--- a/selenium_tests/course_info/course_info_remove_people_tests.py
+++ b/selenium_tests/course_info/course_info_remove_people_tests.py
@@ -46,12 +46,12 @@ class RemovePeopleTests(CourseInfoBaseTestCase):
 
         '''Note: Page needs to reloaded as tests are persistently failing
         for element not found - perhaps cache has changed" for different test
-        id (as the test ID are being removed so DOM has changed. This ensures
+        id (as the test ID are being removed so DOM has changed). This ensures
         that the page is reloaded.'''
         self.driver.refresh()
         self.setUp()
         self._load_test_course()
 
-        # Verifies that user has been removed from course.  
+        # Verifies that user has been removed from course.
         self.assertTrue(
             self.people_page.is_person_removed_from_list(test_user))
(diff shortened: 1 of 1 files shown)


selenium_tests/course_info/course_info_remove_people_tests.py
```diff
@@ -43,5 +43,15 @@ def test_remove_person(self, test_case_id, test_user, canvas_role, role_id):
...
 
         self.assertTrue(self.people_page.delete_was_successful())
+
+        '''Note: Page needs to reloaded as tests are persistently failing
+        for element not found - perhaps cache has changed" for different test
+        id (as the test ID are being removed so DOM has changed). This ensures
+        that the page is reloaded.'''
+        self.driver.refresh()
+        self.setUp()
+        self._load_test_course()
```
elliottyates: I think that to reload the page you don't need the `refresh()`, right? 

In general, I agree that this seems overkill to deal with the stale element exception. I also am not sure where that stale element exception comes from, it would be worth further investigation -- but if this fixes it, let's get it running and worry about efficiency later.

elliottyates:
👍

hktest:
I read up on a bunch of online resources on this stale_element business, the best I found was this: http://docs.seleniumhq.org/exceptions/stale_element_reference.jsp  (basically, because we are removing users from the page, the DOM gets changed so elements become stale) Then again, it's not clear why it's happening mostly on the remove and not add.     This is one of 2 solutions -- another way is supposed to be just re-locating the element (but which I can't get to work).   In any case, yeah, let's see if we can get this running first and then see if we can refactor/fix for efficiency.  Thanks!

PR #109, opened 2016-04-19
Task/sapnamysore/tlt 2119/ordering failures

sapnamysore:
Converted  the alert/warning to a datable and set the default sorting to be on the first column(search term). Disabled other options on the datatable.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -633,8 +633,12 @@
         $scope.setCourseInstance($routeParams.courseInstanceId);
 
         // configure the alert datatable
-        $scope.dtOptionsWarning = {searching: false, paging: false,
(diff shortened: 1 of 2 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -632,6 +632,10 @@
         $scope.selectedRole = $scope.roles[0];
         $scope.setCourseInstance($routeParams.courseInstanceId);
 
+        // configure the alert datatable
+        $scope.dtOptionsWarning = {searching: false, paging: false,
```
rascalking: might just be me, but once it spills over to multiple lines, i usually prefer one-line-per-key for dicts/objects:

```
$scope.dtOptionsWarning = {
    info: false,
    ordering: ([1, 'asc']),
    paging: false,
    searching: false,
};
```

rascalking:
👍

PR #111, opened 2016-04-20
Sprint 46 TLT-2119 Use angular-drf for gets

rascalking:
NOTE: this breaks jasmine horribly, fixes coming in a follow-on PR


elliottyates:
👍

PR #113, opened 2016-04-21
Hchow/2581 selenium tests for cross listing tool

hktest:
This PR is just building the page structure for the new cross-listing tool.
The tests for cross listing will be covered in TLT-2589 (also in this sprint).


selenium_tests/cross_listing/cross_listing_base_test_case.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        super(CrossListingBaseTestCase, cls).setUpClass()
+
+        cls.USERNAME = settings.SELENIUM_CONFIG.get('selenium_username')
+        cls.PASSWORD = settings.SELENIUM_CONFIG.get('selenium_password')
+
+        cls.CANVAS_BASE_URL = settings.SELENIUM_CONFIG.get('canvas_base_url')
+        cls.test_settings = settings.SELENIUM_CONFIG['cross_listing']
```
eparker71: there's a mix of using get('key') with ['key'], I think it's better to use ['key'] as the script will fail immediately if the values are not present, whereas using get will always return 'None' even if no values are present and could to a misdiagnosed failed test

selenium_tests/cross_listing/page_objects/cross_listing_index_page.py
```diff
@@ -0,0 +1,12 @@
+from selenium.webdriver.common.by import By
+
+from selenium_tests.cross_listing.page_objects.cross_listing_base_page_object\
+    import CrossListingBasePageObject
+
+
+class Locators(object):
+    HEADING_ELEMENT = (By.XPATH, '//h3[contains(.,"Cross Listing")]')
```
eparker71: This certainly works but I think getting by ID would be better. I think it's totally reasonable to go ahead and update the template being tested by adding id's. Not something that needs to happen on this, just food for thought.

selenium_tests/cross_listing/cross_listing_base_test_case.py
hktest: Good point - fixed and thanks!

selenium_tests/cross_listing/page_objects/cross_listing_index_page.py
hktest: Great, it's conditioned against xpath right now because there is no ID element to check against the header.  I've requested for this id to be added when folks pick up TLT-1314.  Thanks for the feedback!   
https://jira.huit.harvard.edu/browse/TLT-1314?focusedCommentId=252942&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-252942

eparker71:
👍 Looks Great!

selenium_tests/account_admin/page_objects/account_admin_dashboard_page_object.py
```diff
@@ -41,8 +41,18 @@ def cross_listing_button_is_displayed(self):
 
     def select_course_info_link(self):
         """
-        select the course info link element and click it
+        select the cross list link element and click it
```
elliottyates: This probably should stay 'course info', no?

selenium_tests/cross_listing/page_objects/cross_listing_base_page_object.py
```diff
@@ -0,0 +1,34 @@
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.common.by import By
+
+from selenium_common.base_page_object import BasePageObject
+
+
+class Locators(object):
+    @classmethod
+    def TD_TEXT_XPATH(cls, search_text):
```
elliottyates: We should probably move this (eventually?) into the base page object, since it is used in so many different tests and is general enough to match a `<td>` on any page.

selenium_tests/cross_listing/page_objects/cross_listing_base_page_object.py
```diff
@@ -0,0 +1,34 @@
...
+        return By.XPATH, "//td[normalize-space(text())='{}']".format(
+            search_text)
+
+
+class CrossListingBasePageObject(BasePageObject):
+    def __init__(self, driver):
+        super(CrossListingBasePageObject, self).__init__(driver)
+        self.focus_on_tool_frame()
+
+    def get_cell_with_text(self, search_text):
```
elliottyates: Ditto for this service -- it could be in the base page object (eventually).

selenium_tests/cross_listing/page_objects/cross_listing_index_page.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class IndexPagePageObject(CrossListingBasePageObject):
```
elliottyates: Nitpick -- do we need two `Page`s in the name? Also, is 'index page' sufficiently clear as to what's on the page? Maybe 'ListPageObject'? I dunno. Like I said, nitpick, not that important -- we can trace through the code to see what's what. Might be worth a comment though to explain what this PO is for (the `HEADING_ELEMENT` doesn't really give the reader an idea of what might be on the page -- although I'm sure as the PO is built out for test cases it will be more obvious).

selenium_tests/cross_listing/cross_listing_base_test_case.py
```diff
@@ -0,0 +1,49 @@
...
+
+        cls.USERNAME = settings.SELENIUM_CONFIG['selenium_username']
+        cls.PASSWORD = settings.SELENIUM_CONFIG['selenium_password']
+
+        cls.CANVAS_BASE_URL = settings.SELENIUM_CONFIG['canvas_base_url']
+        cls.test_settings = settings.SELENIUM_CONFIG['cross_listing']
+        cls.TOOL_RELATIVE_URL = cls.test_settings['relative_url']
+        cls.TOOL_URL = urljoin(cls.CANVAS_BASE_URL, cls.TOOL_RELATIVE_URL)
+
+        cls.course_info_parent_page = AccountAdminDashboardPage(cls.driver)
```
elliottyates: This is a little confusing, looks like a cut-n-paste. Maybe we should just call it the account_admin_dashboard_page. :)

selenium_tests/cross_listing/cross_listing_base_test_case.py
```diff
@@ -0,0 +1,49 @@
...
+from selenium_common.pin.page_objects.pin_login_page_object \
+    import PinLoginPageObject
+
+from selenium_tests.account_admin.page_objects.account_admin_dashboard_page_object \
+    import AccountAdminDashboardPage
+from selenium_tests.cross_listing.page_objects.cross_listing_index_page \
+    import IndexPagePageObject
+
+
+class CrossListingBaseTestCase(BaseSeleniumTestCase):
```
elliottyates: A lot of this class shares similar code with the `CourseInfoBaseTestCase`, it would be good to (eventually) make it all DRYer.

canvas_account_admin_tools/settings/local.py
```diff
@@ -64,6 +64,9 @@
          },
       }
    },
+   'cross_listing': {
+      'relative_url': 'accounts/8/external_tools/99'
```
elliottyates: Seems like we could use the `account_admin` `relative_url` for all the tests under the dashboard...? They all start there, right?

selenium_tests/cross_listing/cross_listing_base_test_case.py
elliottyates: Further to [this comment](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/113/files#r60773312), all the tests that start at the dashboard could just use the same `TOOL_URL`.

selenium_tests/account_admin/page_objects/account_admin_dashboard_page_object.py
hktest: Seeing double.  Nice catch, fixed.

selenium_tests/cross_listing/cross_listing_base_test_case.py
hktest: That's what you get for thinking you can save a little time and copy and paste.  :(  
Nice catch; I'll be a little more careful about it.

selenium_tests/cross_listing/page_objects/cross_listing_index_page.py
hktest: It's supposed to model the landing (home) page of the cross-listing tool.  I've modified the page object name to "main_page_object" and added a quick comment on what it's suppose to model.   There is only ONE page to the cross-listing tool right now (so the index page seems intuitive to me...)   The "heading_element" is there is to cover the _is_loaded condition (to check test is in the right place).  I think this PO will be much clearer once more tests are in place.

canvas_account_admin_tools/settings/local.py
hktest: Ah, I read and re-read it this comment like 3 times to figure out how this means (and how to fix).  It makes sense now.  I've removed this cross_listing relative URL from local.py.  Thanks.

selenium_tests/cross_listing/cross_listing_base_test_case.py
hktest: Fixed; removed references to cross-listing relative URL.

selenium_tests/cross_listing/page_objects/cross_listing_base_page_object.py
hktest: Will give it a shot to move this into base object in this round.  This is being used by course_info as well, so it's probably cleaner to move it in sooner than later.   I've only done the update to selenium_common once, so it'll be a good practice as well.

selenium_tests/cross_listing/page_objects/cross_listing_base_page_object.py
hktest: Ditto comment below, will give it a shot move and into base page object for this round.

hktest: Related PR for moving this into selenium common here:
https://github.com/Harvard-University-iCommons/selenium_common/pull/16

selenium_tests/cross_listing/page_objects/cross_listing_base_page_object.py
hktest: Related PR for moving this into selenium common here:
https://github.com/Harvard-University-iCommons/selenium_common/pull/16

selenium_tests/cross_listing/cross_listing_base_test_case.py
hktest: Re: "A lot of this class shares similar code with the CourseInfoBaseTestCase, it would be good to (eventually) make it all DRYer."   I can't easily get it to work, so I will tinker around it with but will see if I can get a working version of it in a later round (not this round).

hktest: Update: Sweet!  Was able to figure it out and get it working to make it more DRYer --- will put in a PR for this or the change will be bundled with TLT-1589.

PR #115, opened 2016-04-25
Created a base_test_case...

hktest:
at the PROJECT level, so all the apps in this project can use the setUpClass. 
All apps in the canvas_account_admin_tools lands on the dashboard.  

This should address the DRY feedback here:  
https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/113#discussion_r60966617

If this is the right approach, other apps in this project can be refactored the same way when we touch those projects. Tested, and it works!

### DIFF
--- a/selenium_tests/canvas_account_admin_tools_base_test_case.py
+++ b/selenium_tests/canvas_account_admin_tools_base_test_case.py
@@ -1,0 +1,29 @@
+from urlparse import urljoin
+
+from django.conf import settings
+
+from selenium_common.base_test_case import BaseSeleniumTestCase
+from selenium_common.pin.page_objects.pin_login_page_object import PinLoginPageObject
+from selenium_tests.account_admin.page_objects.account_admin_dashboard_page_object import AccountAdminDashboardPage
+
+
+class AccountAdminBaseTestCase(BaseSeleniumTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(AccountAdminBaseTestCase, cls).setUpClass()
+
+        cls.USERNAME = settings.SELENIUM_CONFIG['selenium_username']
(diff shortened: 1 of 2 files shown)


selenium_tests/canvas_account_admin_tools_base_test_case.py
```diff
@@ -0,0 +1,29 @@
...
+        cls.PASSWORD = settings.SELENIUM_CONFIG['selenium_password']
+        cls.CANVAS_BASE_URL = settings.SELENIUM_CONFIG['canvas_base_url']
+        cls.TOOL_RELATIVE_URL = settings.SELENIUM_CONFIG['account_admin']['relative_url']
+        cls.TOOL_URL = urljoin(cls.CANVAS_BASE_URL, cls.TOOL_RELATIVE_URL)
+
+        cls.acct_admin_dashboard_page = AccountAdminDashboardPage(cls.driver)
+        cls.acct_admin_dashboard_page.get(cls.TOOL_URL)
+        login_page = PinLoginPageObject(cls.driver)
+        if login_page.is_loaded():
+            login_page.login_xid(cls.USERNAME, cls.PASSWORD)
```
elliottyates: This works for now -- to make it more extensible later we can refactor it. We can create a service on the login_page (eventually!...) that takes a username, password, and login type, thus it can be overridden for HUIDs, XIDs, HarvardKey via email address, etc. Just thinking out loud.

selenium_tests/canvas_account_admin_tools_base_test_case.py
```diff
@@ -0,0 +1,29 @@
...
+        else:
+            print '(User {} already logged in to PIN)'.format(cls.USERNAME)
```
elliottyates: We can probably update this message to not mention PIN...?

elliottyates:
👍 looks great!

PR #117, opened 2016-04-26
Task/sapnamysore/tlt 2119/jasmine confirm add people to course

sapnamysore:
-Made the modalInstance a scope variable in order to close and dismiss the dialog for Jasmine testing(This is similar to how is set in confirmRemove). Also tested the change locally.
-Jasmine tests

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -198,7 +198,7 @@
             var searchTermList = $scope.getSearchTermList(searchTerms);
             // open a modal confirmation box and as the user to verify they want to add
             // the number of users they entered.
-            var modalInstance = $uibModal.open({
+            $scope.modalInstance = $uibModal.open({
                 animation: true,
                 templateUrl: 'partials/add-people-to-course-confirmation.html',
                 controller: function ($scope, $uibModalInstance, numPeople, selectedRoleName) {
@@ -215,7 +215,7 @@
                 }
             });
 
-            modalInstance.result.then(function modalConfirmed() {
+            $scope.modalInstance.result.then(function modalConfirmed() {
                 $scope.addPeopleToCourse(searchTermList);
             });            
         };
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -198,7 +198,7 @@
             var searchTermList = $scope.getSearchTermList(searchTerms);
             // open a modal confirmation box and as the user to verify they want to add
             // the number of users they entered.
-            $scope.modalInstance = $uibModal.open({
+            $scope.confirmAddModalInstance = $uibModal.open({
                 animation: true,
                 templateUrl: 'partials/add-people-to-course-confirmation.html',
                 controller: function ($scope, $uibModalInstance, numPeople, selectedRoleName) {
@@ -215,9 +215,12 @@
                 }
             });
 
-            $scope.modalInstance.result.then(function modalConfirmed() {
+            $scope.confirmAddModalInstance.result.then(function modalConfirmed() {
                 $scope.addPeopleToCourse(searchTermList);
-            });            
+            })
+            .finally(function() {
+                $scope.confirmAddModalInstance = null;
+            });
         };
         $scope.confirmRemove = function(membership) {
             // creates a new remove user confirmation modal, and
--- a/course_info/tests/jasmine/controllers/PeopleControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/PeopleControllerSpec.js
@@ -106,9 +106,64 @@ describe('Unit testing PeopleController', function() {
         it('updates the progress bar regardless of success/failure');
     });
     describe('confirmAddPeopleToCourse', function() {
-        it('reflects the right role and number of people');
-        it('initiates the process if confirmed');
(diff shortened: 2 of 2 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -198,7 +198,7 @@
             var searchTermList = $scope.getSearchTermList(searchTerms);
             // open a modal confirmation box and as the user to verify they want to add
             // the number of users they entered.
-            var modalInstance = $uibModal.open({
+            $scope.modalInstance = $uibModal.open({
```
elliottyates: If we're going to put this on the scope, can we call it something more specific, like `confirmAddModal`?

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -215,7 +215,7 @@
                 }
             });
 
-            modalInstance.result.then(function modalConfirmed() {
+            $scope.modalInstance.result.then(function modalConfirmed() {
```
elliottyates: Do we need a `finally()` here that [removes this from the controller scope](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/task/elliottyates/tlt-2119/course_info_add_multiple_people_jasmine_tests/course_info/static/course_info/js/controllers/PeopleController.js#L252), then, like we do for `$scope.confirmRemoveModalInstance`?

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -181,9 +181,64 @@ describe('Unit testing PeopleController', function() {
 
     });
     describe('confirmAddPeopleToCourse', function() {
-        it('reflects the right role and number of people');
-        it('initiates the process if confirmed');
-        it('doesn\'t initiate the process if canceled');
+        beforeEach(function() {
```
elliottyates: There's now a [one-liner](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/task/elliottyates/tlt-2119/course_info_add_multiple_people_jasmine_tests/course_info/tests/jasmine/controllers/PeopleControllerSpec.js#L39) that might be able to replace this CI mocking stuff:
```
clearInitialCourseInstanceFetch();
```

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -181,9 +181,64 @@ describe('Unit testing PeopleController', function() {
...
+            var ci = {
+                course_instance_id: $routeParams.courseInstanceId,
+                title: 'confirmAddPeopleToCourse test',
+            };
+            courseInstances.instances[ci.course_instance_id] = ci;
+            controller = $controller('PeopleController', {$scope: scope});
+        });
+
+        xit('reflects the right role and number of people', function() {
```
elliottyates: I also have no idea how to do this, after some research. Let's just make it pending, and put some comments in it, maybe something like:
```
// todo: cannot yet find a way to access the modal's controller scope within a test
it('reflects the right role and number of people');
```

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -181,9 +181,64 @@ describe('Unit testing PeopleController', function() {
...
+            expect(modalScope.selectedRoleName).toEqual("Student");
+        });
+
+        it('initiates the process if confirmed', function() {
+            var searchTerm ="123456789, \n987654321";
+            var searchTermList = [
+                ['123456789', ['123456789']],
+                ['987654321 ', ['987654321']],
+            ];
+            spyOn(scope, 'getSearchTermList').and.returnValue(searchTermList);
```
elliottyates: I don't think we need to mock this, really. It's a straightforward call that doesn't rely on anything external. In any case, if you wanted to continue to mock it, the `searchTermList` var we'd expect it to return would be:
```
["123456789", "987654321"]
```

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
```diff
@@ -181,9 +181,64 @@ describe('Unit testing PeopleController', function() {
...
+            spyOn(scope, 'addPeopleToCourse');
+            scope.confirmAddPeopleToCourse(searchTerm);
+            scope.$digest();  // resolves modal instantiation
+
+            //The modalInstance in the controller had to be made a scope variable
```
elliottyates: Since we're going to leave the 'reflects the right role and number of people' test pending, everything above this line can be moved to the `beforeEach()` for these remaining tests.

course_info/static/course_info/js/controllers/PeopleController.js
sapnamysore: Yes, I had it in my earlier version and put it while applying my stashed changes.

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
sapnamysore: Cool

course_info/tests/jasmine/controllers/PeopleControllerSpec.js
sapnamysore: I hadn't mocked it initially, but it was throwing some errors due to other unrelated stuff that has gone away since then.  And the return was a direct dump from Chrome Dev Tools :-) . Cleaning it

PR #121, opened 2016-04-29
tlt-2119: course info add people jasmine tests: small fix to frontend…

elliottyates:
… from PR 118


elliottyates:
This was already PR'd as part of 118, just didn't make it into the original jasmine branch due to user error.

PR #122, opened 2016-05-02
tlt-2589: selenium cross listing tests - fixes for ddt, selectors

elliottyates:


hktest:
Tested.  That missing asterisk, geesh.  Thanks Elliott!

PR #124, opened 2016-05-10
Task/hchow/2589 selenium tests for cross listing tool

hktest:
Tested locally and works.  

Will need to up the version of selenium common first, as this test uses the api methods in selenium common.   
Related branch there:  https://github.com/Harvard-University-iCommons/selenium_common/pull/17


selenium_tests/account_admin/page_objects/account_admin_dashboard_page_object.py
```diff
@@ -10,8 +10,7 @@ class AccountAdminDashboardPageLocators(object):
     # use text() for exact match (should only match on dashboard page)
     PAGE_TITLE = (By.XPATH, '//h3[text()="Admin Console"]')
     COURSE_INFO_LINK = (By.XPATH, "//a[contains(.,'Find Course Info')]")
-    CROSS_LISTING_DIV = (By.ID, "cross-listing-button")
-
+    CROSS_LISTING_DIV = (By.XPATH, "//a[contains(., 'Cross-list Courses')]")
```
elliottyates: Hmm, just curious, why are we no longer looking by ID? (If the text changes later, we'll have to change this PO locator as well...)

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @data(*get_xl_data(TEST_DATA_CROSS_LISTING_MAPPINGS))
+    @unpack
+    def test_add_cross_listing_pairing(self, test_case,
+                                       primary_cid,
+                                       secondary_cid,
+                                       expected_result,
+                                       expected_text):
+        """
+        TLT-2589, AC #1 and 7
```
elliottyates: Are we linking with the manual test cases, the selenium ticket, and/or the original story? Seems like all of the above would be helpful in some way.

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+        This test adds a valid cross-list pairing to the cross-listing table
+        """
+        # This adds the pairing to the cross-list table and clicks submit
```
elliottyates: Should we attempt to remove the crosslisting first, to ensure any clutter from previously failed tests don't cause this test to fail? (Similar to what we're doing in the remove tests)

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+        self.main_page.add_cross_listing_pairing(primary_cid, secondary_cid)
+
+        # This verifies that the confirmation box appears
+        self.assertTrue(
```
elliottyates: This could be rolled into the `add_cross_listing_pairing()` (i.e. that wouldn't return until the confirmation alert shows up).

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+                self.main_page.confirm_presence_of_confirmation_alert()
+        )
+
+        # Verifies a successful cross-list add
+        if expected_result == 'success':
```
elliottyates: Should we be paranoid and make this
```
if expected_result == 'success':
    ...
elif expected_result == 'fail':
    ...
else:
    raise RuntimeError('error in test data bla bla')
```

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+
+            #  Verifies successful add confirmation
+            actual_text = self.main_page.get_actual_confirmation_text()
+            expected_text = self.main_page.get_expected_confirmation_text(
+                        expected_text)
+            self.assertEqual(actual_text, expected_text,
```
elliottyates: I'm confused about why we need to do all this. Can we just assert that the locator is found for the expected text? (See comments on the PO code itself)

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -0,0 +1,55 @@
...
+            # Clean up and remove the cross-listed course when test is done
+            self.api.remove_xlisted_course(primary_cid, secondary_cid)
+
+        #  Verifies an unsuccessful cross-list
+        if expected_result == 'fail':
+            actual_text = self.main_page.get_actual_confirmation_text()
+            expected_text_on_page = \
+                self.main_page.get_expected_confirmation_text(expected_text)
+            self.assertEqual(actual_text, expected_text_on_page,
+                             "Error: the error message locator cannot be "
```
elliottyates: Ditto about a simpler assertion.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+    @classmethod
+    def DELETE_CROSSLIST_ICON(cls, data_xlist_map_id):
+        """ returns a locator for the xlist link for the xlist_map_id"""
+        return By.CSS_SELECTOR, "a[data-xlist-map-id='{}']".format(
+                data_xlist_map_id)
+
+    @classmethod
+    def ERROR_TEXT_LOCATOR(cls, expected_text):
+        """ returns a locator for expected error text"""
+        return By.XPATH, ".//div[contains(.,'{}') and " \
```
elliottyates: nice

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+        primary_cid_field = self.find_element(*Locators.PRIMARY_CID_ADD_FIELD)
+        primary_cid_field.send_keys(primary_cid)
+
+        # Fills in the secondary CID field for cross-listing
+        secondary_cid_field = self.find_element(
+                *Locators.SECONDARY_CID_ADD_FIELD)
+        secondary_cid_field.send_keys(secondary_cid)
+
+        # Clicks on submit button to pair the cross listing
+        self.find_element(*Locators.SUBMIT_BUTTON).click()
```
elliottyates: In future iterations of our tests, we might want to put the `confirm_presence_of_confirmation_alert()` logic directly into these services -- in other words, the service call won't return until there's some indication that it worked. But that's not necessary now, and I expect it'll be complicated, so we can discuss the merits and gotchas offline. :) Just throwing it out there as an idea.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+
+    def confirm_presence_of_confirmation_alert(self):
+        """
+        :param self:
+        :return: Confirms that an alert box returns after add
```
elliottyates: Maybe "Confirms that an alert box appears/is present, e.g. after an add or delete action"...? -- as the service itself doesn't know or have anything to do with add, specifically.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+        except TimeoutException:
+            return False
+        return True
+
+    def delete_cross_listing_pairing(self, data_xlist_map_id):
+        """ Deletes cross-listing pairing through crosslisting tool in
+        admin console and confirms delete in modal window"""
+        self.find_element(*Locators.DELETE_CROSSLIST_ICON(
+                data_xlist_map_id)).click()
+        self.find_element(*Locators.DELETE_MODAL_CONFIRM).click()
```
elliottyates: Presumably this is not on the page at all until you click on the DELETE_CROSSLIST_ICON? If it's on the page, but hidden, and this line runs before it's visible, we might get errors. But if you're consistently _not_ getting errors, should be ok.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+
+    def get_actual_confirmation_text(self):
```
elliottyates: Maybe `get_confirmation_message_text()`? The 'actual' makes me nervous! :)

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+        """
+        Returns the confirmation text after add
```
elliottyates: This isn't really connected just to the add functionality -- the delete also shows confirmation text (and other functions in the future might as well). Services are independent of each other, as written.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -2,15 +2,89 @@
...
+        """
+        alert = self.find_element(*Locators.CONFIRMATION_ALERT)
+        confirmation_text = alert.text.strip()
+        return confirmation_text
+
+    def get_expected_confirmation_text(self, expected_text):
```
elliottyates: Hmm, so this is a bit confusing to me -- if we're matching something with that text, why return the same text? Why not just make this a boolean like `is_confirmation_message()`?

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
+from ddt import ddt, data, unpack
+
+from selenium_common.base_test_case import get_xl_data
+
+from selenium_tests.cross_listing.cross_listing_base_test_case import \
+    CrossListingBaseTestCase, TEST_DATA_CROSS_LISTING_MAPPINGS
```
elliottyates: I wonder if, in the future, this could just be a property of the base test case class, so it can be accessed via `cls.test_data_xls`... just thinking out loud. It's fine the way it is (and has been all along), just more verbose this way.

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+
+@ddt
+class RemoveCrossListingTests(CrossListingBaseTestCase):
+
+    @data(*get_xl_data(TEST_DATA_CROSS_LISTING_MAPPINGS))
+    @unpack
+    def test_remove_mapping(self, test_case_id, primary_cid,
```
elliottyates: Do we want to link these with the test cases/ticket(s) as well?

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+                            secondary_cid, expected_result,
+                            expected_alert_text):
+
+        """
+        Note: ICOMMONS_REST_API_HOST environment needs to match the LTI tool
```
elliottyates: We can remove this comment - the backend isn't changing the cache at all for these. This was true for the people tests, but isn't relevant for these tests.

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+        environment (because of shared cache interactions)
+
+        1. remove existing cross-list mappings via rest api first to
+        avoid incidental conflict when we try to add a new mapping.
+        2. add via cross-listing via rest api that we want to remove
```
elliottyates: first 'via' can be removed, I think

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+            self.api.remove_xlisted_course(primary_cid, secondary_cid)
+
+            # Add the xlisted pair via rest api
+            self.api.add_xlisted_course(primary_cid, secondary_cid)
+
+            # Get the xlist_map_id so look for the delete icon
+            xlist_map_id = self.api.lookup_xlist_map_id(primary_cid,
+                                                        secondary_cid)
+
+            # Reload cross_listing_page due to changes from REST API Add.
```
elliottyates: Maybe ...?
```
# Reload cross listing main page; REST API calls changed data, and if we're on the cross listing page already then the datatable won't necessarily show the most recent state of the data until we reload the table.
```

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+            if not self.acct_admin_dashboard_page.is_loaded():
+                self.acct_admin_dashboard_page.get(self.TOOL_URL)
+            # navigate to cross-list tool
+            self.acct_admin_dashboard_page.select_cross_listing_link()
+
+            # Clicks on the delete icon associated with the xlist_map_id
```
elliottyates: Note: this will only work if the data shows up on the first page... not a problem now, but in a term of two our test data may need to get updated. Might want to call that out in a comment or something.

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -0,0 +1,56 @@
...
+            self.main_page.delete_cross_listing_pairing(xlist_map_id)
+
+            # Verifies pair has been de-crosslisted by confirmation message
+            expected_text = "Successfully de-cross-listed {} and {}.".format(
+                                                    primary_cid, secondary_cid)
+            actual_text = self.main_page.get_actual_confirmation_text()
+            self.assertEqual(actual_text, expected_text,
+                             "Expected success message is '{}' but message is "
```
elliottyates: I think that assertEqual does something like this automatically, so not sure you need this extra message.
```
class TestEqual(unittest.TestCase):
    def test_example(self):
        a = 'a'
        b = 'b'
        self.assertEqual(a, b)
```
running:
```
======================================================================
FAIL: test_example (__main__.TestEqual)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "<stdin>", line 5, in test_example
AssertionError: 'a' != 'b'

----------------------------------------------------------------------
```

elliottyates:
👍 some places that could be cleaned up, but otherwise looks good!

selenium_tests/account_admin/page_objects/account_admin_dashboard_page_object.py
hktest: Because that element is a really a link. I totally get what you're saying and had initially set it up initially to look it up by id as well but that had returned an element not found a handful of times (unexpectedly).  I haven't run into the same issue using the anchor tag instead, so I decided to opt for it instead for this case in favor of dealing with flakiness.  Can always change it, if need be.

selenium_tests/cross_listing/add_cross_listing_tests.py
hktest: Good point, and fixed.  I am opting to put in the original requirement Jira #, related selenium subtask Jira #, and the ACs (which doesn't get changed) the test cover.  Probably won't be linking to specific test case # since that goes through quite a bit of changes.

selenium_tests/cross_listing/add_cross_listing_tests.py
hktest: Yes, makes sense!  Will add in that logic.

selenium_tests/cross_listing/add_cross_listing_tests.py
hktest: Broke it out this way because I thought it would be cleaner to have methods that does very specific things.  But since there is no return of True/False in the add_cross_listing_pairing() and it's mostly a simulation of entering of ID and clicking, this suggestion fits well.  Fixing.

selenium_tests/cross_listing/add_cross_listing_tests.py
hktest: Yes, thanks for this discussion.  Modifying  to check on "is_locator_text_present" instead, not on actual specific text.

selenium_tests/cross_listing/add_cross_listing_tests.py
hktest: Also fixed thanks.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: thanks, future iterations. :)  already integrated this into the "add crosslisting" logic per previous PR comment.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: no longer an issue.  method is integrated in add per earlier PR feedback.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: Yes, true.  This clicking on modal window hasn't been an issue from all the runs i've done.   I'll keep an eye out for errors and will add an explicit wait if it's an issue.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: no longer relevant. this method has been deleted.  checking by locator text instead as discussed.  thanks!

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: ditto comment above.  no longer relevant.

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: ditto above.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Will punt this one for a future round. The base test case class for cross listing looks pretty clean, and it's clear where the test data is coming from.  The benefit is true as you stated and the add tests could make use for that as well.  For this round, let's keep as is - it currently works with same effect.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Fixed.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Thanks - I thought this is true anytime there is a call to modify data via rest api.  Removed the comment, but not really sure when changing the cache happens.  Will ping you offline to pick your brains on this one. :)  Thanks!

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Fixed.  Also modified text to be clearer.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Fixed.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Done.  Called out in comment with possible options.

selenium_tests/cross_listing/remove_cross_listing_tests.py
hktest: Fixed!

PR #125, opened 2016-05-11
(Sprint 47 leftovers) tlt-1314: xlisting - unit tests

elliottyates:
Tests for proxy inserting current user information when creating xlist maps.

We should ensure, once merged, that this is actually tested via https://jenkins.tlt.harvard.edu/view/QA%20Jobs/job/canvas_account_admin_tools-run-unit-tests/ (should be picked up automatically).

### DIFF
--- a/canvas_account_admin_tools/tests.py
+++ b/canvas_account_admin_tools/tests.py
@@ -73,11 +73,20 @@ class CanvasAccountAdminToolsAPIProxyTests(TestCase):
         Proxy should not add last_modified_by to non-xlistmap POST data
         """
 
+        # test_path represents the path fragment handed to
+        # views.icommons_rest_api_proxy from the router (see the <path> arg
+        # in urls.py). If it doesn't contain xlist_map, the proxy layer
+        # should not add user auditing info
+        test_path = 'api-path/something_else/'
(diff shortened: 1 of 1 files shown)


canvas_account_admin_tools/tests.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_proxy_no_xlist_maps_auditing(self, mock_proxy_view, *args, **kwargs):
+        """
+        Proxy should not add last_modified_by to non-xlistmap POST data
+        """
+
+        test_post_data = {"any": "thing"}
+        self._setup_request(method='POST', data=json.dumps(test_post_data))
+
+        # call the view
+        response = icommons_rest_api_proxy(self.request, self.request.path_info)
```
sapnamysore: path_info would be the fake-path?

sapnamysore:
👍

canvas_account_admin_tools/tests.py
elliottyates: Yes! :) It doesn't really matter in this case, as long as it isn't xlist_map... I guess I should code that explicitly in the test so it doesn't bite us later though. Thanks!

PR #128, opened 2016-05-16
Attempting to fix Jenkins error by adding an explicit wait to check alert appears

hktest:
Adding in explicit wait to make sure the result_message element appears first before checking for text.  This is to fix this error:  https://jenkins.tlt.harvard.edu/job/canvas_account_admin_tools_selenium_tests/125/console

Not getting this error when running the add_cross_listing_tests individually (before and now).  However, I reran the full regression suite for this project and the failed test also passed, so I'm hopeful this should do it.

### DIFF
--- a/selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
+++ b/selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
@@ -69,7 +69,7 @@ class MainPageObject(CrossListingBasePageObject):
 
     def is_locator_text_present(self, expected_text):
         """
-        Check to that locator text appears on the confirmation text box
+        Check that the locator text appears in the confirmation text box
         """
         try:
             WebDriverWait(self._driver, 60).until(lambda s: s.find_element(
--- a/selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
+++ b/selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
@@ -69,9 +69,11 @@ class MainPageObject(CrossListingBasePageObject):
 
     def is_locator_text_present(self, expected_text):
         """
-        Check to that locator text appears on the confirmation page
+        Check to that locator text appears on the confirmation text box
         """
         try:
+            WebDriverWait(self._driver, 60).until(lambda s: s.find_element(
+                *Locators.CONFIRMATION_ALERT).is_displayed())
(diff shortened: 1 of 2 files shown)


selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -69,9 +69,11 @@ def delete_cross_listing_pairing(self, data_xlist_map_id):
 
     def is_locator_text_present(self, expected_text):
         """
-        Check to that locator text appears on the confirmation page
+        Check to that locator text appears on the confirmation text box
```
sapnamysore: Nitpick: 'Check that the locator text..'? :-)

hktest: Good nitpick and fixed!

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -69,9 +69,11 @@ def delete_cross_listing_pairing(self, data_xlist_map_id):
 
     def is_locator_text_present(self, expected_text):
         """
-        Check to that locator text appears on the confirmation page
+        Check that the locator text appears in the confirmation text box
         """
         try:
+            WebDriverWait(self._driver, 60).until(lambda s: s.find_element(
```
sapnamysore: Is 60 the minimum that works? Have we tried with 30? In any case, either works, but want to check if we can reduce.

sapnamysore:
👍

hktest:
Per feedback, the link to the actual test report (with the failure that this PR is intended to fix) is here: https://jenkins.tlt.harvard.edu/job/canvas_account_admin_tools_selenium_tests/ws/reports/canvas_account_admin_tools_test_report_20160514_03_15_48.html

selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
hktest: Good point.  I can lower to 30s, but let's try with 60s first, because I want to make sure that this actually fixes the failure -- in case Jenkins was just slow and couldn't find the element without the explicit waits (since this hadn't been an issue when I ran the file individually).  Can totally revisit the length of waits.  Thanks!

PR #131, opened 2016-06-03
Task/sapnamysore/tlt 2558/add course info flags

sapnamysore:
PR contains changes for https://jira.huit.harvard.edu/browse/TLT-2558 . Also tweaked a font style per Vittorio's suggestion so it will wrap properly in the  new Canvas UI


elliottyates:
👍

PR #133, opened 2016-06-06
Modified Jamine test to add  the new flags.

sapnamysore:


course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -211,7 +214,9 @@ describe('Unit testing DetailsController', function () {
             expect(dc.courseInstance['xlist_status']).toEqual('N/A');
             expect(dc.courseInstance['description']).toEqual('<p>hello</p>');
             // members.count is not defined it should equal 0
-            expect(dc.courseInstance['members']).toEqual(100);
```
elliottyates: Any reason why this went away, but the comment stayed?

elliottyates:
One comment about possibly accidental removal of an existing check/test, otherwise 👍

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
sapnamysore: Thanks, accidental deletion when inserting new code. Adding it back

PR #134, opened 2016-06-07
Update test logic for non-editable course 

hktest:
Updating an existing test due to business logic/UI changes in TLT-2640. 

Business Logic change: the Submit and Reset buttons now appear for non-editable (registrar fed courses, not just ILE/SB courses) in Course Details page.   

Also some light code cleanup. 
Tested and ran successfully.

### DIFF
--- a/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
+++ b/selenium_tests/course_info/course_info_details_noneditable_course_tests.py
@@ -16,8 +16,8 @@ class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
         verify non-SB/ILE courses cannot be edited
         """
         # Verify that buttons to edit page are not present
-        self.assertFalse(self.detail_page.
-                         verify_buttons_to_edit_page_are_present())
+        self.assertTrue(self.detail_page.
+                        verify_buttons_to_edit_page_are_present())
 
         non_editable_fields = [
         'course_instance_id',
@@ -34,6 +34,5 @@ class CourseInfoDetailsNonEditTests(CourseInfoBaseTestCase):
 
         for element in non_editable_fields:
             #  verify that the field is not rendered as an input element
-            self.assertFalse(
-                    self.detail_page.is_element_displayed_as_input_field(element))
-
+            self.assertFalse(self.detail_page.
+                             is_element_displayed_as_input_field(element))
--- a/selenium_tests/course_info/page_objects/course_info_detail_page_object.py
+++ b/selenium_tests/course_info/page_objects/course_info_detail_page_object.py
@@ -53,9 +53,9 @@ class CourseInfoDetailPageObject(CourseInfoBasePageObject):
 
     def verify_buttons_to_edit_page_are_present(self):
         try:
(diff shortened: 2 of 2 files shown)


selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -51,19 +51,12 @@ def go_to_people_page(self):
...
-        except NoSuchElementException:
-            return False
-        return True
-
     def verify_buttons_to_edit_page_are_present(self):
-
-        if (self.is_locator_element_present(Locators.RESET_FORM_BUTTON) and
-            self.is_locator_element_present(Locators.SUBMIT_FORM_BUTTON)):
+        try:
+            (self.find_element(*Locators.RESET_FORM_BUTTON).is_displayed()
```
eparker71: I think you need an if here?

eparker71:
This looks great, I just had one comment about a missing "if" good to go once that's accounted for. 👍

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
hktest: Great, nice catch!  I thought the  try/except would replace the need for the "if" in this case (since if either element is not displayed it fails out "Except" and return False.  But it's probably much cleaner to add in the "if".  Fixed and thanks!

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
```diff
@@ -16,8 +16,8 @@ def test_regular_course_not_editable(self):
         verify non-SB/ILE courses cannot be edited
         """
         # Verify that buttons to edit page are not present
```
elliottyates: (Might want to update this comment and the test docstring, plus line 9 above)

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -51,20 +51,13 @@ def go_to_people_page(self):
...
 
-    def is_locator_element_present(self, locator_element):
+    def verify_buttons_to_edit_page_are_present(self):
         try:
-            self.find_element(locator_element)
+            if (self.find_element(*Locators.RESET_FORM_BUTTON).is_displayed()
+            and self.find_element(*Locators.SUBMIT_FORM_BUTTON).is_displayed()):
+                return True
         except NoSuchElementException:
             return False
```
elliottyates: Don't you need another `return False` at the bottom to ensure that if either isn't displayed, but both are found, you are still returning `False`?

selenium_tests/course_info/course_info_details_noneditable_course_tests.py
hktest: Fixed!

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
hktest: Yes, thanks for the tip on using return instead of potentially returning 2 False (one in if/then statement, and the other in the NoSuchElementException).  Updated as well - thanks!

PR #135, opened 2016-06-13
TLT-2588 set course title to short title or 'Untitled Course' if titles are undefined

eparker71:
Set the course title to short_title if there is no title. 
Set the course title to 'Untitled Course' if there is no short title. 
Added Jasmine tests.

### DIFF
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -110,14 +110,24 @@
                 });
 
         };
-
+        dc.getCourseDescription = function(course) {
+            // If a course's title is [NULL], attempt to display the short title.
+            // If the short title is also [NULL], display [School] 'Untitled Course' [Term Display]
+            if(typeof course.title != "undefined" && course.title.trim().length > 0){
+                return course.title;
+            }
+            else if(typeof course.short_title != "undefined" && course.short_title.trim().length > 0){
+                return course.short_title;
+            }
+            return 'Untitled Course';
+        };
         dc.getFormattedCourseInstance = function (ciData) {
             // This is a helper function that formats the raw CourseInstance
             // API response data for display in the UI
             var ci = ciData;  // shorten for brevity, preferable to `with()`
             var courseInstance = {};
             if (ci) {
-                courseInstance['title'] = ci.title;
+                courseInstance['title'] = dc.getCourseDescription(ci);
                 courseInstance['school'] = ci.course.school_id.toUpperCase();
                 courseInstance['term'] = ci.term.display_name;
                 courseInstance['year'] = ci.term.academic_year;
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -286,6 +286,18 @@
             // return the filtered list
             return filteredResults;
         };
+
+        $scope.getCourseDescription = function(course) {
+            // If a course's title is [NULL], attempt to display the short title.
+            // If the short title is also [NULL], display [School] 'Untitled Course' [Term Display]
+            if(typeof course.title != "undefined" && course.title.trim().length > 0){
+                return course.title;
+            }
+            else if(typeof course.short_title != "undefined" && course.short_title.trim().length > 0){
+                return course.short_title;
+            }
+            return 'Untitled Course';
+        };
         $scope.getFormattedCourseInstance = function(ci) {
             // This is a helper function that formats the CourseInstance metadata
             // and is combination of existing logic in
@@ -293,7 +305,7 @@
             // render functions.
             var courseInstance = {};
             if (ci) {
-                courseInstance['title']= ci.title;
+                courseInstance['title']= $scope.getCourseDescription(ci);
                 courseInstance['school'] = ci.course ?
                         ci.course.school_id.toUpperCase() : '';
                 courseInstance['term'] = ci.term ? ci.term.display_name : '';
--- a/course_info/static/course_info/js/controllers/SearchController.js
+++ b/course_info/static/course_info/js/controllers/SearchController.js
@@ -108,10 +108,10 @@
             $scope.getCourseDescription = function(course) {
                 // If a course's title is [NULL], attempt to display the short title.
                 // If the short title is also [NULL], display [School] 'Untitled Course' [Term Display]
-                if(typeof course.title != "undefined" && course.title.length > 0){
+                if(typeof course.title != "undefined" && course.title.trim().length > 0){
                     return course.title;
                 }
-                else if(typeof course.short_title != "undefined" && course.short_title.length > 0){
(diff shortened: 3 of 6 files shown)


course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -105,12 +105,24 @@
                 });
             };
 
+            $scope.getCourseDescription = function(course) {
+                // If a course's title is [NULL], attempt to display the short title.
+                // If the short title is also [NULL], display [School] 'Untitled Course' [Term Display]
+                if(typeof course.title != "undefined" && course.title.length > 0){
```
elliottyates: Probably a good idea to trim() the value before checking the length, in case it is a string of blank spaces (which won't display a link).

course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -105,12 +105,24 @@
...
+                    return course.title;
+                }
+                else if(typeof course.short_title != "undefined" && course.short_title.length > 0){
+                    return course.short_title;
+                }
+                return 'Untitled Course';
```
elliottyates: The way I read the AC, we need to prepend the school and append the term display, is that not the way it was meant?

course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -105,12 +105,24 @@
...
+            };
+
             $scope.courseInstanceToTable = function(course) {
                 // This logic is reused in PeopelController.getFormattedCourseInstance
                 // So any changes to data format need to be done in both places.
                 // TODO: Eventually move the reusable logic to a separate library
                 var cinfo = {};
-                cinfo['description'] = course.title;
+                cinfo['description'] = $scope.getCourseDescription(course);
```
elliottyates: As per the method docstring above, don't we need to update the PeopleController as well?

course_info/tests/jasmine/controllers/SearchControllerSpec.js
```diff
@@ -90,4 +90,30 @@ describe('Unit testing SearchController', function() {
...
+            expect(result).toEqual('short');
+        });
+
+        it('should display the [school_id] Untitled Course [term dsiplay name] when no title or short title are present', function(){
+            var course = { title: '', short_title: ''};
+            result = scope.getCourseDescription(course);
+            expect(result).toEqual('Untitled Course');
+        });
+
+        it('should display Untitled Course when no data is present present', function(){
```
elliottyates: double 'present'

course_info/tests/jasmine/controllers/SearchControllerSpec.js
```diff
@@ -90,4 +90,30 @@ describe('Unit testing SearchController', function() {
...
+        it('should dipslay the course short title if title is not present and short title is', function() {
+            var course = { title: '', short_title: 'short'};
+            result = scope.getCourseDescription(course);
+            expect(result).toEqual('short');
+        });
+
+        it('should display the [school_id] Untitled Course [term dsiplay name] when no title or short title are present', function(){
+            var course = { title: '', short_title: ''};
+            result = scope.getCourseDescription(course);
+            expect(result).toEqual('Untitled Course');
```
elliottyates: This isn't checking the same thing as the description on line 107...?

course_info/tests/jasmine/controllers/SearchControllerSpec.js
```diff
@@ -90,4 +90,30 @@ describe('Unit testing SearchController', function() {
...
+
+    describe('getCourseDescription()', function() {
+        it('should display the course title if title is present', function () {
+            var course = { title: 'ABC'};
+            result = scope.getCourseDescription(course);
+            expect(result).toEqual('ABC');
+        });
+
+        it('should dipslay the course short title if title is not present and short title is', function() {
+            var course = { title: '', short_title: 'short'};
```
elliottyates: Should probably also check that if both title and short_title are present, title is preferred.

course_info/static/course_info/js/controllers/SearchController.js
eparker71: I thought that, but then I realized that the two columns on either side of title are School and Term. I did make an assumption that that's what they meant. I will double check thought, you might be right.

course_info/static/course_info/js/controllers/SearchController.js
eparker71: yeah, good call!

course_info/static/course_info/js/controllers/SearchController.js
eparker71: hmm, yeah, I didn't see that. Good catch!

course_info/static/course_info/js/controllers/SearchController.js
elliottyates: Hmm, yeah, good point - that's confusing! [Term Display] isn't quite what's exactly the next column over, but that info is all there in the table, right? So you're probably right, we don't need that stuff -- worth double-checking (and updating any test descriptions, and maybe specifying that in the ticket for clarity). Thanks!

course_info/tests/jasmine/controllers/SearchControllerSpec.js
eparker71: yup, added the appropriate test

eparker71: yup, added that test

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -110,14 +110,24 @@
                 });
 
         };
-
+        dc.getCourseDescription = function(course) {
```
elliottyates: Hmm, yes, it looks like `getFormattedCourseInstance()` is called from `handleCourseInstanceResponse()`, so for ILEs this may change the behavior of the form, as we discussed offline -- let's think about what we want this page to do. Maybe just let it stay blank? Seems ok to me, in this case. A comment describing why `title` differs here from the search page might be helpful, though.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -286,14 +286,26 @@
             // return the filtered list
             return filteredResults;
         };
+
+        $scope.getCourseDescription = function(course) {
```
elliottyates: Yeah I guess this could go either way -- it's less important on the people page as it's just in that info box at the top, and we already know what context we're in by the course instance ID, for example. But probably nice to have it in the box, so I guess leave it?

elliottyates:
Once we figure out whether we want this functionality for the course detail page/remove the code and tests for that page, looks good. 👍

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: Discussed with Chrissy and the ILE and SB courses should also display 'Untitled Course' when no title exists.

PR #141, opened 2016-06-20
Task/eparker71/tlt 2570/backend allow add

eparker71:
Hook up the front end to the backend. The front end already had mock hooks for these calls. I just used these points to hook in to the real backend.

### DIFF
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -38,22 +38,26 @@
 
         dc.associateNewSite = function() {
             dc.associateNewSiteInProgress = true;
-            // todo: call backend
-            $timeout(function() {
-                var fakeCourse = {
+            data = {
+              'external_id': dc.newCourseSiteURL
+            };
+            var post_course_site_url = djangoUrl.reverse(dc.apiProxy,
+                        [dc.apiBase + dc.courseInstance.course_instance_id + '/sites/']);
+            $http.post(post_course_site_url, data).then(function(response){
+                var new_ourse = {
                     course_site_url: dc.newCourseSiteURL,
                     map_type: 'official'
                 };
-                dc.courseInstance.sites.push(fakeCourse);
+                dc.courseInstance.sites.push(new_ourse);
                 dc.newCourseSiteURL = '';
                 dc.associateNewSiteInProgress = false;
-                // dc.alerts.form.siteOperationFailed = {
-                //     show: true,
-                //     operation: 'associating',
-                //     // todo: response.statusText || 'None'
-                //     details: 'None'};
-            },
-                2000
+            }, function(response) {
+                dc.handleAjaxErrorResponse(response);
+                dc.alerts.form.siteOperationFailed = {
+                    show: true,
+                    operation: 'associating',
+                    details: response.statusText || 'None'};
+                }
             );
         };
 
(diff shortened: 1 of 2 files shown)


sapnamysore:
👍

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -38,22 +38,27 @@
...
+                var new_course = {
                     course_site_url: dc.newCourseSiteURL,
-                    map_type: 'official'
+                    map_type: 'official',
+                    site_map_id: response.data.site_map_id
                 };
-                dc.courseInstance.sites.push(fakeCourse);
+                dc.courseInstance.sites.push(new_course);
                 dc.newCourseSiteURL = '';
                 dc.associateNewSiteInProgress = false;
```
elliottyates: Does this need to be in the equivalent of an `always` block, or at least reset in the failure branch as well?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -72,35 +77,39 @@
...
                     }
                 }
             });
+            dc.confirmDissociateSiteModalInstance.result.then(
+                function modalConfirmed() {
+                    var delete_course_site_url = djangoUrl.reverse(dc.apiProxy,
+                        [dc.apiBase + dc.courseInstance.course_instance_id + '/sites/' + dc.courseInstance.sites[siteListIndex].site_map_id + '/']);
+                    $http.delete(delete_course_site_url)
+                        .then(function(response){
+                            dc.dissociateSiteInProgressIndex = siteListIndex;
```
elliottyates: Doesn't this need to be set at the beginning of is function, not after making the ajax call?

course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -72,35 +77,39 @@
...
+                            dc.courseInstance.sites.splice(siteListIndex, 1);
+                            dc.dissociateSiteInProgressIndex = null;
```
elliottyates: Does this need to be in the equivalent of an `always` block, or at least reset in the failure branch as well?

course_info/templates/course_info/partials/details.html
```diff
@@ -261,7 +261,6 @@ <h3 class="breadcrumbs">
                       <a href="{{site.course_site_url}}" target="_blank">{{site.course_site_url}}</a>
                     </div>
                     <div class="col-sm-3">
-                      <!-- todo: switch $index to site_id when available -->
                       <button type="button" class="btn btn-default pull-right"
                               ng-click="dc.dissociateSite($index)">
```
elliottyates: Wouldn't it be better to use a map of sites than an index-based list, so that if multiple operations happened we wouldn't rely on the reliability of the indexes staying the same (whereas the `site_map_id` will be unique)?

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: I don't think I understand what you mean. I think we only want to push a new site onto the list if the post was successful? no?

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: do you mean line 102?

course_info/static/course_info/js/controllers/DetailsController.js
eparker71: ok, yeah, that makes sense. Will add the finally block with dc.dissociateSiteInProgressIndex = null;

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: Yes, because the ['in progress' indicator on the template](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/task/elliottyates/tlt-2570/course_details_add_remove_course_sites/course_info/templates/course_info/partials/details.html#L266) is linked to this -- but if we only set it here, then it won't ever show (or it will flicker very quickly) as here we're already finished with the operation. So I think it should be before line 100, no?

course_info/static/course_info/js/controllers/DetailsController.js
elliottyates: (discussing offline)

course_info/templates/course_info/partials/details.html
eparker71: so, using the index is nice because you get the object in one step 'dc.courseInstance.sites[siteListIndex].site_map_id' doing a lookup seems unnecessary here, I may be missing something though.

elliottyates: Index could change though while something is querying the backend, I would think -- for instance, we might delete site 1 and site 2, and if site 1 is removed first, the index to site 2 might no longer be valid/reliable. So that's why I think a `site_map_id`-keyed object would be better. That's why the comment was there, sorry for not being clearer initially!

PR #142, opened 2016-06-20
Sprint 50: TLT-2490 - people tool jasmine tests

elliottyates:

PR #143, opened 2016-06-21
Task/eparker71/tlt 2570/jasmine tests

eparker71:
Final PR for 2570, contains updated and new jasmine tests

### DIFF
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -685,16 +685,15 @@ describe('Unit testing DetailsController', function () {
         it('should make the delete call to disasscociate a url with the course instance', function(){
             //delete by specifying the index
             siteListIndex = 1;
+            var siteURL = '';
+            var site_map_id = '';
             deleteURL ='/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args=api%2Fcourse%2Fv2%2Fcourse_instances%2F1234567890%2Fsites%2F999%2F';
 
             dc.dissociateSite(siteListIndex);
-
-            dc.confirmDissociateSiteModalInstance.close();
-            scope.$digest();
-
+            scope.$digest();// resolves modal
+            dc.confirmDissociateSiteModalInstance.close(siteURL, site_map_id);
             $httpBackend.expectDELETE(deleteURL).respond(204, {});
-            //$httpBackend.flush(1);
-
+            $httpBackend.flush(1);
             expect(dc.courseInstance.sites.length).toEqual(1);
         });
 
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -608,12 +608,14 @@ describe('Unit testing DetailsController', function () {
                     {
                         external_id: 'https://x.y.z/888',
                         site_id: '888',
-                        map_type: 'official'
+                        map_type: 'official',
+                        site_map_id: '888'
                     },
                     {
                         external_id: 'https://x.y.z/999',
                         site_id: '999',
-                        map_type: 'unofficial'
+                        map_type: 'unofficial',
+                        site_map_id: '999'
                     }
                 ]
             };
@@ -628,11 +630,11 @@ describe('Unit testing DetailsController', function () {
             dc.newCourseSiteURL = 'http://testing.com';
             dc.associateNewSite();
             $httpBackend.expectPOST(postSiteUrl)
-                .respond(201, JSON.stringify({status: "success"}));
+                .respond(201, JSON.stringify({site_map_id: '777'}));
             $httpBackend.flush(1);
-            siteList = [{external_id: 'https://x.y.z/888', site_id: '888', map_type: 'official'},
-                        {external_id: 'https://x.y.z/999', site_id: '999', map_type: 'unofficial'},
-                        {course_site_url: 'http://testing.com', map_type: 'official', site_map_id: undefined}
+            siteList = [{external_id: 'https://x.y.z/888', site_id: '888', map_type: 'official', site_map_id: '888'},
+                        {external_id: 'https://x.y.z/999', site_id: '999', map_type: 'unofficial', site_map_id: '999'},
+                        {course_site_url: 'http://testing.com', map_type: 'official', site_map_id: '777'}
                         ];
             expect(dc.courseInstance.sites.length).toEqual(3);
             expect(dc.courseInstance.sites).toEqual(siteList);
--- a/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
+++ b/course_info/tests/jasmine/controllers/DetailsControllerSpec.js
@@ -13,6 +13,14 @@ describe('Unit testing DetailsController', function () {
         '=api%2Fcourse%2Fv2%2Fcourse_instances%2F' + courseInstanceId +
         '%2Fpeople%2F&-source=xreg_map';
 
+    var deleteSiteUrl =
+        '/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args' +
+        '=api%2Fcourse%2Fv2%2Fcourse_instances%2F' + courseInstanceId + '%2Fsites%2F999%2F';
+
+    var postSiteUrl =
+        '/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args' +
+        '=api%2Fcourse%2Fv2%2Fcourse_instances%2F' + courseInstanceId + '%2Fsites%2F';
+
     // set up the test environment
     beforeEach(function () {
         // load the app and the templates-as-module
@@ -619,9 +627,8 @@ describe('Unit testing DetailsController', function () {
 
         it('should make the post call to associate a new site with the course instance', function(){
             dc.newCourseSiteURL = 'http://testing.com';
-            post_url = '/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args=api%2Fcourse%2Fv2%2Fcourse_instances%2F1234567890%2Fsites%2F';
             dc.associateNewSite();
-            $httpBackend.expectPOST(post_url)
+            $httpBackend.expectPOST(postSiteUrl)
                 .respond(201, JSON.stringify({status: "success"}));
             $httpBackend.flush(1);
             siteList = [{external_id: 'https://x.y.z/888', site_id: '888', map_type: 'official'},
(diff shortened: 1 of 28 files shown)


course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -13,6 +13,14 @@ describe('Unit testing DetailsController', function () {
         '=api%2Fcourse%2Fv2%2Fcourse_instances%2F' + courseInstanceId +
         '%2Fpeople%2F&-source=xreg_map';
 
+    var deleteSiteUrl =
```
elliottyates: In the future since we're bringing the djangoUrl module into our test framework we can probably just use its `reverse()` function to make the code more concise (see #142). No need right now.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -13,6 +13,14 @@ describe('Unit testing DetailsController', function () {
...
+        '/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args' +
+        '=api%2Fcourse%2Fv2%2Fcourse_instances%2F' + courseInstanceId + '%2Fsites%2F999%2F';
```
elliottyates: This hardcodes the site ID -- in other tests we're adding it as part of the test. Just something to be aware of, that the other URLs here are generic/'base'-type URLs.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
         });
     });
+
+    describe('associateNewSiteHandleKey', function(){
+        it('should call associateNewSite when a user hits the enter key', function(){
+            spyOn(dc, 'validNewSiteURL').and.returnValue(true);
+            spyOn(dc, 'associateNewSite');
+            var aEvent = angular.element.Event('keydown');
+            aEvent.keyCode = 13;
+            $("input").trigger(aEvent);
```
elliottyates: Neat. Learned something new here.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            dc.associateNewSiteHandleKey(aEvent);
+            expect(dc.validNewSiteURL).toHaveBeenCalled();
+            expect(dc.associateNewSite).toHaveBeenCalled();
+
+        });
+
+        it('should not call associateNewSite when a user enteres an invalid url', function(){
+            spyOn(dc, 'validNewSiteURL').and.returnValue(false);
+            spyOn(dc, 'associateNewSite');
```
elliottyates: I guess a few of these lines could be `beforeEach()` and `afterEach()` for DRY-ness, but that may be overkill - with only two tests, it does read clearly this way.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+                sites: [
+                    {
+                        external_id: 'https://x.y.z/888',
+                        site_id: '888',
+                        map_type: 'official'
+                    },
+                    {
+                        external_id: 'https://x.y.z/999',
+                        site_id: '999',
+                        map_type: 'unofficial'
```
elliottyates: These look different from the sites in the `dissociateSite`, should they be consistent?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+                    }
+                ]
+            };
+            dc = $controller('DetailsController', {$scope: scope});
```
elliottyates: These lines could also be in a function that's called where needed for DRYness, but that isn't necessary for the PR.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+                .respond(500, JSON.stringify({status: "invalid url"}));
+            $httpBackend.flush(1);
+            expect(dc.alerts.form.siteOperationFailed).toEqual({show: true, operation: 'associating', details: 'None'});
+        });
+
+    });
+
+    describe('dissociateSite', function(){
+        beforeEach(function () {
+            var ci = {
```
elliottyates: Seems like much of this could be DRYer as well -- but I recognize some people prefer the test data as part of the main `describe()` block, even if there's lots of duplication.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+        });
+
+        it('should make the delete call to disasscociate a url with the course instance', function(){
+            //delete by specifying the index
+            siteListIndex = 1;
+            var siteURL = '';
+            var site_map_id = '';
+
+            dc.dissociateSite(siteListIndex);
+            scope.$digest();// resolves modal
```
elliottyates: `dc.$digest()` ?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            dc.confirmDissociateSiteModalInstance.close(siteURL, site_map_id);
```
elliottyates: huh, this is indeed where I thought you had to `dc.$digest()` again.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            $httpBackend.expectDELETE(deleteSiteUrl).respond(204, {});
+            $httpBackend.flush(1);
+            expect(dc.courseInstance.sites.length).toEqual(dc.courseInstance.sites.length);
```
elliottyates: What is this doing? Don't we want to check that it equals `original count - 1` ?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+        });
+
+        it('doesn\'t delete any sites if the process if canceled', function() {
+            dc.dissociateSite(siteListIndex);
```
elliottyates: Is `siteListIndex` undefined here? A little confusing, maybe a comment or an explicit `undefined` might be clearer.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            scope.$digest()
```
elliottyates: `dc.$digest()` ?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            dc.confirmDissociateSiteModalInstance.dismiss()
```
elliottyates: Ditto -- has anything actually changed at this point?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            expect(dc.courseInstance.sites.length).toEqual(2);
+        });
+
+        it('should show an alert if the delete fails', function(){
+            //delete by specifying the index
+            siteListIndex = 1;
+            var siteURL = '';
+            var site_map_id = '';
+            dc.dissociateSite(siteListIndex);
+            scope.$digest();// resolves modal
```
elliottyates: `dc.$digest()` ?

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            dc.confirmDissociateSiteModalInstance.close(siteURL, site_map_id);
+            $httpBackend.expectDELETE(deleteSiteUrl).respond(404, {});
+            $httpBackend.flush(1);
+            expect(dc.alerts.form.siteOperationFailed.show).toEqual(true);
```
elliottyates: Hmm, interesting that this works without another `dc.$digest()` to resolve the `dc.confirmDissociateSiteModalInstance.result`.

people_tool/tests/jasmine/SearchControllerTests.js
```diff
@@ -1,31 +1,26 @@
 describe('Unit testing people_tool SearchController', function() {
     var $controller, $rootScope, $routeParams, $compile, djangoUrl,
-        $httpBackend, $log, $templateCache, $window;
+        $httpBackend, $log, $templateCache, $timeout, $window;
```
elliottyates: Not sure how these commits got in there, but it'll come out in the wash.

elliottyates:
A few suggestions and tweaks, but nothing major other than double-checking that the `dissociateSite` tests are testing what we want them to. Barring that, 👍 .

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
sapnamysore: sorry, yes, I had it checking for 1(which is original count -1) and must have accidentally replaced both sides with same variable. Fixed

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
sapnamysore: Yes, this is the second digest  I mentioned that I had to remove. It complains about  a digest already being in progress

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
sapnamysore: Have initialized the index, not sure why it din't throw an error before. And validating no change in count and also added a isUndefined check

sapnamysore: Further debugging showed siteListIndex set to 1 when it is not declared, so looks like the scope of a var is beyond just the function it is declared in.

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: they don't have to be for the test to work, but they prob should be for sanity sake :)

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
```diff
@@ -559,4 +566,161 @@ describe('Unit testing DetailsController', function () {
...
+            $httpBackend.expectGET(courseInstanceURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.expectGET(peopleURL).respond(200, JSON.stringify({status: "success"}));
+            $httpBackend.flush(2);
+        });
+
+        it('should make the post call to associate a new site with the course instance', function(){
+            dc.newCourseSiteURL = 'http://testing.com';
+            dc.associateNewSite();
+            $httpBackend.expectPOST(postSiteUrl)
+                .respond(201, JSON.stringify({status: "success"}));
```
eparker71: yup, we should, I'll add that return

course_info/tests/jasmine/controllers/DetailsControllerSpec.js
eparker71: yeah, these tests were added to an existing spec, so while it does make sense to refactor much of the spec, that will have to be a tech debt item.

PR #146, opened 2016-06-24
Reintroducing the model and fixtures and removing the migration that …

sapnamysore:
…deleted the 'ExternalTool' : in order to fix unit tests that  were failing cute to the cleanup. May handle this separately as a tech debt.


elliottyates:
👍

PR #147, opened 2016-06-29
Task/sapnamysore/tlt 2073/add person details table

sapnamysore:
This PR contains changes for adding the selected person's details to the Courses page . Added a service to share data between the 2 controllers. 
TODO: Needs to add the huid/xid badge to the person table

### DIFF
--- a/people_tool/static/people_tool/css/main.css
+++ b/people_tool/static/people_tool/css/main.css
@@ -5,6 +5,12 @@
     pointer-events: none;
 }
 
+
+table.person-info-table {
+    display: table;
+    width: 920px;
+}
+
 table.dataTable thead .sorting,
 table.dataTable thead .sorting_asc,
 table.dataTable thead .sorting_desc,
--- a/people_tool/static/people_tool/js/controllers/SearchController.js
+++ b/people_tool/static/people_tool/js/controllers/SearchController.js
@@ -164,13 +164,11 @@
                     dataType: 'json'
                 }).done(function dataTableGetDone(data, textStatus, jqXHR) {
                     $scope.messages = [];
-                    console.log(" done ");
                     callback({
                         recordsTotal: data.count,
                         recordsFiltered: data.count,
                         data: data.results
                     });
-                    console.log(data.results);
                 })
                 .fail(function dataTableGetFail(data, textStatus, errorThrown) {
                     $log.error('Error getting data from ' + url + ': '
--- a/people_tool/static/people_tool/js/controllers/SearchController.js
+++ b/people_tool/static/people_tool/js/controllers/SearchController.js
@@ -2,7 +2,7 @@
     var app = angular.module('PeopleTool');
     app.controller('SearchController', SearchController);
 
-    function SearchController($scope, $compile, djangoUrl, $log, $timeout) {
+    function SearchController($scope, $compile, djangoUrl, $log, $timeout, personInfo) {
         // tracks state of interactive datatable elements
         $scope.datatableInteractiveElementTabIndexes = [];
         $scope.messages = [];
@@ -98,9 +98,23 @@
             // can still display a clickable anchor link
             var lastName = (full.name_last.trim().length > 0) ?
                 full.name_last : '(none)';
-            return '<a href="' + '#/people/' + full.univ_id + '/courses/">'
+            return '<a href="' + '#/people/' + full.univ_id + '/courses/" ' +
+                'ng-click="setSelectedPersonInfo(\''+full.name_last+'\',\''+
+                full.name_first+'\','+full.univ_id+',\''+full.email_address+'\''+',\''+full.role_type_cd+'\')">'
                 + lastName + '</a>';
         };
+        $scope.setSelectedPersonInfo =
+            function(name_last, name_first, univ_id, email, role_type_cd){
+                //Populate the  personInfo object  with the selected person's details
+                // to be used in the next screen
+            personInfo.details = {
+                name_last: name_last,
+                name_first: name_first,
+                univ_id: univ_id,
+                email_address: email,
+                role_type_cd: role_type_cd
+            };
+        }
         $scope.searchPeople = function(event) {
             if ($scope.queryString.length > $scope.searchType.maxLength) {
                 $scope.messages = [{
@@ -150,11 +164,13 @@
                     dataType: 'json'
                 }).done(function dataTableGetDone(data, textStatus, jqXHR) {
                     $scope.messages = [];
+                    console.log(" done ");
                     callback({
                         recordsTotal: data.count,
                         recordsFiltered: data.count,
                         data: data.results
                     });
+                    console.log(data.results);
                 })
                 .fail(function dataTableGetFail(data, textStatus, errorThrown) {
                     $log.error('Error getting data from ' + url + ': '
--- a/people_tool/static/people_tool/js/services/AngularDRFService.js
+++ b/people_tool/static/people_tool/js/services/AngularDRFService.js
@@ -1,5 +1,0 @@
-(function() {
-    var app = angular.module('PeopleTool');
-
-    app.factory('angularDRF', ['$http', '$q', angularDRF]);
-})();
--- a/people_tool/static/people_tool/js/services/AngularDRFService.js
+++ b/people_tool/static/people_tool/js/services/AngularDRFService.js
@@ -1,0 +1,5 @@
+(function() {
+    var app = angular.module('PeopleTool');
+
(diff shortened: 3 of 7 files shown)


people_tool/static/people_tool/js/controllers/SearchController.js
```diff
@@ -98,9 +98,23 @@
             // can still display a clickable anchor link
             var lastName = (full.name_last.trim().length > 0) ?
                 full.name_last : '(none)';
-            return '<a href="' + '#/people/' + full.univ_id + '/courses/">'
+            return '<a href="' + '#/people/' + full.univ_id + '/courses/" ' +
+                'ng-click="setSelectedPersonInfo(\''+full.name_last+'\',\''+
+                full.name_first+'\','+full.univ_id+',\''+full.email_address+'\''+',\''+full.role_type_cd+'\')">'
```
eparker71: I was looking at how to do this in angular and it looks like this is a place where we could use a service to pass the data from one controller to another. Now that we have a working version with this code, wanna try creating a service? We could do some XP coding?

elliottyates: Or perhaps instead of an external <a> link, we can simply use `ng-click` to call an angular controller function which sets the person in the service and changes `$location` to load the `#/people/univ_id/courses/` route.

elliottyates: (it would clean up this link immensely and put all the logic in one place, in a JS controller function -- might be easier to read that way)

people_tool/static/people_tool/css/main.css
```diff
@@ -5,6 +5,12 @@
     pointer-events: none;
 }
 
+
+table.person-info-table {
```
elliottyates: Just curious, why are we doing this? Could we have a comment as to what this accomplishes?

people_tool/static/people_tool/js/controllers/SearchController.js
```diff
@@ -98,9 +98,23 @@
...
                 + lastName + '</a>';
         };
+        $scope.setSelectedPersonInfo =
+            function(name_last, name_first, univ_id, email, role_type_cd){
```
elliottyates: I wonder if there's not some way to just pass this the row's `full` object, which contains all these things in all the right keys, to reduce explicit param passing. Maybe the service code itself can selectively remember the keys it cares about. Takes some of this stuff out of the controller and moves it into the service, which would be a good practice for reusability...? I may be missing some limitations with that, though.

people_tool/static/people_tool/js/services/AngularDRFService.js
```diff
@@ -0,0 +1,5 @@
+(function() {
```
elliottyates: I don't believe we need this file in the people_tool. Is there a reason this is included?

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
     {% verbatim %}
     <small><i class="fa fa-chevron-right"></i></small>
     <a href="#/">Search People</a>
-    <small><i class="fa fa-chevron-right"></i></small>
-    {{ personId }}
+
+    <span ng-if="selectedPersonInfo.name_first">
```
elliottyates: Could this fail in cases where the person has a blank first name? We might want a different if clause here and/or a fallback for the name_last, name_first text below (perhaps in a controller function or something).

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
...
+    </span>
     <span ng-if="operationInProgress"><i class="fa fa-refresh fa-spin"></i></span>
   </h3>
 </nav>
 
 <main>
+ <div ng-if='selectedPersonInfo' class="row">
+
+  <div class="col-md-12">
+      <div id="personInfoDT_wrapper" class="dataTables_wrapper form-inline no-footer">
```
elliottyates: Why are we using the `.form-line` here?

people_tool/static/people_tool/js/controllers/SearchController.js
sapnamysore: @eparker71 : so within the setSelectedPersonInfo, the person metadats is being used to build a person object and I do use a service to pass it from one controller to another. [https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/task/sapnamysore/tlt-2073/add_person_details_table/people_tool/static/people_tool/js/services/PersonInfoService.js#L1]

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
...
+              <td class="sorting" tabindex="0">{{selectedPersonInfo.name_first}}</td>
+              <td class="sorting" tabindex="0">{{selectedPersonInfo.univ_id}}</td>
+              <td class="sorting" tabindex="0">{{selectedPersonInfo.email_address}}</td>
+            </tr>
+          </tbody>
+        </table>
+      </div>
+  </div>
+</div>
+<p></p>
```
elliottyates: Seems like we could use an ID-tagged or class-tagged div for the above block and introduce a bottom margin to get the spacing we want, vs using a blank semantic tag to get things to display the way we want.

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
...
+  <div class="col-md-12">
+      <div id="personInfoDT_wrapper" class="dataTables_wrapper form-inline no-footer">
 
+        <table id="personInfoDT"
+               class="table table-striped table-hover ng-isolate-scope dataTable no-footer person-info-table"
+               role="grid" aria-describedby="personInfoDT_info"
+               cellspacing="0">
+          <thead>
+            <tr role="row">
+                <th tabindex="0" id="lname" aria-controls="DataTables_Table_0" rowspan="1" colspan="1" aria-sort="ascending">Last Name</th>
```
elliottyates: Do we need the id, aria-controls, and aria-sort properties here?

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
...
 </nav>
 
 <main>
+ <div ng-if='selectedPersonInfo' class="row">
+
+  <div class="col-md-12">
+      <div id="personInfoDT_wrapper" class="dataTables_wrapper form-inline no-footer">
 
+        <table id="personInfoDT"
+               class="table table-striped table-hover ng-isolate-scope dataTable no-footer person-info-table"
```
elliottyates: Do we need the `.ng-isolate-scope`? And is the `.no-footer` appropriate both on this table element and the wrapper div above this?

people_tool/templates/people_tool/partials/courses.html
```diff
@@ -6,14 +6,46 @@ <h3 class="breadcrumbs">
...
+               role="grid" aria-describedby="personInfoDT_info"
```
elliottyates: Doesn't look like an element with `#personInfoDT_info` exists on the page, should we remove this property or add that element?

people_tool/static/people_tool/css/main.css
elliottyates: Ok sorry this is clearer now from what's below (but I'm curious why we don't do a combination of `.col-offset` and `col-*` like we do for the main table and for things like the search progress bar on other pages)?

people_tool/static/people_tool/js/controllers/SearchController.js
sapnamysore: I tried passing the row's full object as a first pass - doesn't work :-( . Hence the passing in as separate params..

people_tool/static/people_tool/js/controllers/SearchController.js
sapnamysore: And yes, it was an attempt to get a simple  working version out , we can definitely cleanup to make it easier to read and maintain.

people_tool/static/people_tool/js/services/AngularDRFService.js
sapnamysore: Thanks, copy and paste issue. It's gone

people_tool/static/people_tool/css/main.css
sapnamysore: I moved the styles from the table to the css element. Didn't attempt using col-offset , but can lookup the search table.

PR #151, opened 2016-07-12
Task/eparker71/tlt 2657/fix duplicate records

eparker71:
updated front end to use the query param include='type_id'. when this param is present the rest api will automatically collapse the rows by role_type_cd so the front end does not need to do this anymore. This has the benefit of allowing the result set to be distinct by user_id. This change corresponds to the backend change located here: https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/92

@elliottyates @sapnamysore


people_tool/static/people_tool/js/controllers/CoursesController.js
```diff
@@ -108,8 +106,8 @@
             }
             return 'Untitled Course';
         };
-        $scope.getProfileRoleTypeCd = function(profile) {
-            return profile ? profile.role_type_cd : '';
+        $scope.getProfileIdType = function(profile) {
```
elliottyates: Looks like we can completely remove this, I can't see anywhere it's used in this app (it came from the course_info app and may not have been used in this one).

people_tool/static/people_tool/js/controllers/SearchController.js
```diff
@@ -86,12 +86,10 @@
             return 'Untitled Course';
         };
         $scope.getProfileRoleTypeCd = function(profile) {
```
elliottyates: Ditto, see https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/151/files#r70500699.

elliottyates:
Only one little thing. When user jumps over to course details (the course_info / Search Course app) and returns we no longer have the info from the PersonInfoService, so we have to update the background call that fetches the person info in the CoursesController as well.

screen shot 2016-07-12 at 3 15 16 pm

elliottyates:
Otherwise, looks good! 👍

PR #152, opened 2016-07-12
added css to remove the border from the alert-pane panel, renamed id …

eparker71:
…for both divs as they were the same. make the names specific to the page and descriptive.


eparker71:
I'm going to go ahead and merge this minor change.

PR #155, opened 2016-07-25
reordered dropdown menu alphabetically and made 'guest' the default r…

eparker71:
…ole.


eparker71:
This PR doesn't really have any code changes, just reordering a static list and setting a default.

hktest:
Looks good!  Confirmed with Eric that  $scope.selectedRole = $scope.roles[4]; is the fourth role in the list as it's ordered, which is "Guest".  :+1:

PR #159, opened 2016-07-26
Task/eparker71/tlt 2669/add course section column

eparker71:


people_tool/static/people_tool/js/controllers/CoursesController.js
```diff
@@ -84,7 +84,7 @@
                 cinfo['code'] = '';
                 cinfo['school'] = '';
             }
-
+            cinfo['section'] = course.section ? course.section : '';
```
hktest: A little nit-picky, but maybe there's a reason, but can this line (line 87) be moved right after line 67, where other similar parameters seemed to be set?

hktest:
Other than the question/comment above, :+1:

people_tool/static/people_tool/js/controllers/CoursesController.js
eparker71: sure

PR #160, opened 2016-07-26
Task/elliottyates/tlt 2668/course details reorg

eparker71:


eparker71:
The original PR for this came from @elliottyates. Two addition stories TLT-2701 and TLT-2699 have been added/merged into this branch as they related to the UI and would have caused merge conflicts. They have already been PR'd.

eparker71:
👍 applies to @elliottyates PR

eparker71:
not addressed in this PR:
1) getCourseDescription() is duplicated across all the controllers. Would be great to make it a module/service.

2) course instance fetch code and course membership code is also duplicated everywhere. Would be great to make these services, then we could share them across the three controllers (details, people, and the new 'sites' controller). That would be DRYer, plus then you wouldn't have to reload the people count as you switched tabs.

hktest:
- Eric PR Elliott's code (which is the bulk of the code here).  
- Went over some of Eric's changes in this PR with Eric.  
- Eric will integration test this when it lands in dev.  
- Note:  some of the refactoring are not addressed in PR (see updated task in trello board) 
- Will ACP extensively, but :+1:  on this PR.

PR #162, opened 2016-07-28
Task/hchow/2668 fix selenium tests due to ui changes

hktest:
Due to intro of new tabs in the Course Details page, some 15 or so existing selenium tests were failing.

- updated locator to use the id for the new people tab.  
- also fixed up a failing test logic (Reset button is no longer on the page). 

to fix the bulk of the failed tests. 

Note that there are still 2 tests that are failing, as I can't yet figure out a way to have selenium deal with dom changes after a toggle from Edit -> Submit, which causes the field to go from a span to input field.  Tests was previously searching on input field (which used to work), because there was no toggle and we simply used the input_field locator.  To be looked into further separately, and possibly to be filed as tech debt.


eparker71:
👍 Looks great

PR #163, opened 2016-07-28
fixed bug where the recently added or deleted url was causing an erro…

eparker71:
…r in the js. in the case where the external_id is undefined, use the course_site_url instead.


hktest:
:+1:

PR #164, opened 2016-08-05
tlt-2603: search people - adding people to concluded course

elliottyates:
If users cannot be added to canvas because the course is concluded, mark those adds as successes and show a specific note to the user to indicate a possible delay in syncing the canvas enrollment list.


eparker71:
👍 Looks great!!!

PR #165, opened 2016-08-16
updated admin console course search add people to pull roles from Can…

eparker71:
…vas. also removed shopper role. fixed some unit tests.


canvas_account_admin_tools/settings/base.py
```diff
@@ -267,4 +267,6 @@
 PERMISSION_PEOPLE_TOOL = 'people_tool'
 PERMISSION_XLIST_TOOL = 'cross_listing'
 
-
+# in search courses, when you add a person to a course. This list
+# controls which roles show up in the drop down
```
elliottyates: Aren't these actually `UserRole` ids? (not Canvas role ids?) Either way, we can be more specific about that in the comment.

course_info/data_migrations/canvas_account_admin_perms.sql
```diff
@@ -13,7 +13,7 @@ INSERT INTO lti_permissions_ltipermission (permission, school_id, canvas_role, a
 	('account_admin_tools', '*', 'Librarian', FALSE),
 	('account_admin_tools', '*', 'ObserverEnrollment', FALSE),
 	('account_admin_tools', '*', 'SchoolLiaison', TRUE),
-	('account_admin_tools', '*', 'Shopper', FALSE),
+	('account_admin_tools', '*', 'Prospective Enrollee', FALSE),
```
elliottyates: I just took this out entirely, as an absence of a rule defaults to `FALSE`. I think it's the same, really. (I also removed `Harvard-Viewer` too)

course_info/templates/course_info/index.html
```diff
@@ -45,6 +45,7 @@
     <script>
       window.globals.STATIC_URL = '{% settings_value "STATIC_URL" %}';
       window.schoolOptions = {{ schools | safe }};
+      window.roles = {{ canvas_roles | safe }};
```
elliottyates: Are these actually Canvas role ids, or `UserRole` ids?

course_info/views.py
```diff
@@ -37,6 +39,27 @@ def partials(request, path):
...
+def _get_canvas_roles():
+    """
+    Get the list of canvas roles from the canvas root account ('self') and
+    create a list in the format the front end wants.
+    :return:
+    """
+    roles = []
+    try:
+        canvas_roles = canvas_api_helpers_roles.get_roles_for_account_id('self')
+        for role_id in canvas_roles:
```
elliottyates: Since it's a dict, I think we can do something like `for role_id, role in canvas_roles.items()`, and then line 54 can read
```
'roleName': role['label']
```

people_tool/migrations/0001_initial.py
```diff
@@ -18,7 +18,7 @@
     ('people_tool', '*', 'Librarian', False),
     ('people_tool', '*', 'ObserverEnrollment', False),
     ('people_tool', '*', 'SchoolLiaison', False),
-    ('people_tool', '*', 'Shopper', False),
+    ('people_tool', '*', 'Prospective Enrollee', False),
```
elliottyates: ditto on whether to include this (and Harvard Viewer) at all

course_info/views.py
```diff
@@ -37,6 +39,27 @@ def partials(request, path):
...
+            if role_id in settings.ADD_PEOPLE_TO_COURSE_ALLOWED_ROLES_LIST:
+                role = {'roleId': role_id,
+                        'roleName': canvas_roles[role_id]['label']}
```
elliottyates: I would prefer `roleLabel`, as it's clearer how it maps to the Canvas API data

course_info/views.py
```diff
@@ -25,6 +26,7 @@ def index(request):
     # to the lti launch's user.
     context = {
         'schools': _get_schools_context(canvas_user_id),
+        'canvas_roles': _get_canvas_roles(),
```
elliottyates: Ditto for `UserRole` vs Canvas role question -- I think that our app uses `UserRole` ids, not Canvas role IDs. So we have to do a mapping using the `UserRole.canvas_role_id` property to connect the right labels to the right `UserRole` ids.

course_info/templates/course_info/index.html
eparker71: This is not just the list of id's, it is a list of dicts that contain id and role label. I can change this to say user_roles instead of canvas_roles.

course_info/views.py
```diff
@@ -37,6 +40,26 @@ def partials(request, path):
...
+    :return:
+    """
+    roles = []
+    try:
+        canvas_roles = canvas_api_helpers_roles.get_roles_for_account_id('self')
+        user_roles = UserRole.objects.filter(pk__in=settings.ADD_PEOPLE_TO_COURSE_ALLOWED_ROLES_LIST)
+        for role in user_roles:
+            label = canvas_roles[role.canvas_role_id]['label']
+            roles.append({'roleId': role.role_id,'roleName': label})
+    except CanvasAPIError:
```
elliottyates: Maybe we just want to `except:`, which would catch `UserRole` lookup and dict key lookup errors as well. Otherwise this `CanvasAPIError` except block should just be right under the `get_roles_for_account_id()` line. The broader `except` might be handy, though. The `logger.exception()` message in that case should also be broader.

elliottyates:
👍

PR #167, opened 2016-08-24
tlt 2670: cross listing search

elliottyates:
users can filter the cross-listed courses list by school, course, term, or course instance

### DIFF
--- a/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
+++ b/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
@@ -7,14 +7,19 @@
 
     function ListController($scope, $http, $timeout, $document, $window,
                             $compile, djangoUrl, $log, $q, $uibModal) {
+        $scope.datatableInteractiveElementTabIndexes = [];
         // expected $scope.message format:
         //   {alertType: 'bootstrap alert class', text: 'actual message'}
         $scope.message = null;
         // operationInProgress can be 'add' or 'remove' -- the distinction is
         // made so that the submit button for the add doesn't show a spinning
         // progress icon while the modal is processing a remove request
-        $scope.operationInProgress = null;
+        $scope.operationInProgress = false;
+        $scope.queryString = '';
         $scope.rawFormInput = {primary: '', secondary: ''};
+        // the datatable emptytable message can be deceiving on first load, so
+        // do not show datatable until data loaded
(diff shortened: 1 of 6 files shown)


cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
```diff
@@ -110,14 +121,87 @@
...
+                $scope.dtInstance.reloadData();
+            }, 0);
+        };
+        $scope.setOperationInProgress = function(newSetting) {
+            // disable datatable and show progress bar if newSetting is truthy
+            $timeout(function() {
+                // notify UI to start/stop showing in-progress messaging
+                $scope.operationInProgress = newSetting;
+                // enable/disable interactive data table elements
+                $scope.setDataTableInteraction(!newSetting);
```
sapnamysore: An  operation is being passed into setOperationInProgress and not a boolean [$scope.setOperationInProgress('search')]. So the value of 'newSetting' would be 'search'.  But looks like setDataTableInteraction is expecting a boolean? I am a bit confused  as setDataTableInteraction seems to be expecting to work with a toggle flag (boolean)?

elliottyates: Right -- it treats any string (the name of the operation) as a boolean for the purposes of determining whether or not to allow interaction with the datatable. That is, if it is a non-zero-length string, Javascript considers it truthy. Perhaps it would be best to explicitly cast it to a boolean using `!!newSetting` first? Or just some clarification in an inline comment?

elliottyates: (to be more concrete, if the `newSetting` string is e.g. `'add'`, then `newSetting` is truthy, and `!newSetting` is `false`, which disables the datatable interaction).

sapnamysore: Thanks for clarifying Elliott! Makes more sense now. I do like the idea of explicitly casting it to a boolean..

elliottyates: will do

sapnamysore:
👍 ( looks like u have a fix for the progress bar continuing to appear after an alert is displayed - please feel free to merge after that fix)

sapnamysore:
👍

PR #168, opened 2016-08-25
tlt-2670: cross_list_courses - jasmine tests for searching

elliottyates:
Since `$scope.search()` is kicked off right away in the controller, mocking it and clearing initial calls to it so that all other tests are in the expected initial state is a little complicated. In the future we should consider wrapping any initial calls a controller needs to make in a service so they can be mocked before we instantiate the controller.

### DIFF
--- a/cross_list_courses/tests/jasmine/controllers/ListControllerSpec.js
+++ b/cross_list_courses/tests/jasmine/controllers/ListControllerSpec.js
@@ -7,17 +7,33 @@ describe('Unit testing ListController', function () {
         '/angular/reverse/?djng_url_name=icommons_rest_api_proxy&djng_url_args' +
         '=api%2Fcourse%2Fv2%2Fxlist_maps%2F';
 
+    /* Helpers */
 
-    function clearInitialxlistFetch() {
-        // ignore the initial template fetch
+    function clearTemplateFetch() {
         $httpBackend.expectGET("partials/list.html").respond(200, '');
         $httpBackend.flush(1);
     }
-
+    function runTimeout() {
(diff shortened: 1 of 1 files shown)


cross_list_courses/tests/jasmine/controllers/ListControllerSpec.js
```diff
@@ -217,49 +234,98 @@ describe('Unit testing ListController', function () {
...
             scope.removeCrosslisting(1, 1, 2);
 
-            expect(scope.operationInProgress).toBe('remove');
+            expect(scope.setOperationInProgress).toHaveBeenCalledWith('remove');
             expect(scope.message).toBe(null);
 
             deleteCrosslistingDeferred.resolve('success!');
             scope.$digest();
 
-            expect(scope.operationInProgress).toBeNull();
```
sapnamysore: Wouldn't we still want to check that the operationInProgress is false after a successful remove?
i.e. ` expect(scope.operationInProgress).toBe(false);`

sapnamysore:
Other than a minor comment, looks good 👍

cross_list_courses/tests/jasmine/controllers/ListControllerSpec.js
elliottyates: The actual `scope.operationInProgress` modifications are now wrapped inside `scope.setOperationInProgress()` and `scope.dtInstance.reloadData`, both of which we've mocked, so the value won't changed in this test case. Instead we'll have to assume that those two methods did their jobs... it's pretty convoluted, isn't it.

PR #169, opened 2016-08-25
tlt-2670: cross_list_courses - selenium test updates

elliottyates:
fixes existing selenium tests, adds search capability test

@hktest @sapnamysore

### DIFF
--- a/selenium_tests/cross_listing/add_cross_listing_tests.py
+++ b/selenium_tests/cross_listing/add_cross_listing_tests.py
@@ -11,11 +11,11 @@ class AddCrossListingTests(CrossListingBaseTestCase):
 
     @data(*get_xl_data(TEST_DATA_CROSS_LISTING_MAPPINGS))
     @unpack
-    def test_add_cross_listing_pairing(self, test_case,
+    def test_add_cross_listing_pairing(self, unused_test_case_id,
                                        primary_cid,
                                        secondary_cid,
                                        expected_result,
-                                       expected_text):
+                                       expected_alert_text):
         """
         Jira requirement story: TLT-1314
         Selenium sub-task: TLT-2589.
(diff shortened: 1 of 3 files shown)


selenium_tests/cross_listing/page_objects/cross_listing_main_page.py
```diff
@@ -35,11 +38,19 @@ def ERROR_TEXT_LOCATOR(cls, expected_text):
 class MainPageObject(CrossListingBasePageObject):
     page_loaded_locator = Locators.HEADING_ELEMENT
 
+    # todo: consider moving into the BasePageObject, with override flexibility
+    def __init__(self, driver):
```
hktest: Understood how you are using it, but not understanding the __init__reference here. Will ping you later. :)

hktest:
just a question I'll take offline. but :+1,  thanks!

selenium_tests/cross_listing/remove_cross_listing_tests.py
```diff
@@ -24,6 +26,7 @@ def test_remove_mapping(self, test_case_id, primary_cid,
         avoid incidental conflict when we try to add a new mapping.
         2.  Add cross-listing via rest api of the pairing that we
         want to test the 'remove' functionality on.
+        3.  Find this new cross-listing using the search box
         3.  Test remove by clicking on the cross-list remove icon via UI
```
sapnamysore: nitick: increment 3 to 4?

selenium_tests/cross_listing/add_cross_listing_tests.py
```diff
@@ -11,44 +11,46 @@ class AddCrossListingTests(CrossListingBaseTestCase):
 
     @data(*get_xl_data(TEST_DATA_CROSS_LISTING_MAPPINGS))
     @unpack
-    def test_add_cross_listing_pairing(self, test_case,
+    def test_add_cross_listing_pairing(self, unused_test_case_id,
```
sapnamysore: Curious about the name 'unused_test_case_id'?

elliottyates: Yes -- unfortunately (?) python doesn't allow the use of multiple throwaway `_` variables in the same function signature, so one recommendation I ran across was to prefix with `unused_`. But apparently [PyLint doesn't like that](http://stackoverflow.com/questions/10107350/how-to-handle-the-pylint-message-idw0612-unused-variable), so maybe this should ultimately be changed to something like `_unused_`. Probably not an important update for this PR, but a thought for the future.

selenium_tests/cross_listing/remove_cross_listing_tests.py
elliottyates: yep

PR #170, opened 2016-08-26
tlt-2650: tool permission simplification

elliottyates:
Uses updated icommons_common.lti_permissions.
- default permissions pulled from lti_permissions (schools and roles) for each app
- includes migration for default permissions

This PR is currently to gather feedback and calibrate so we can update the other apps in a consistent way after tweaking this approach.
@sapnamysore


canvas_account_admin_tools/requirements/base.txt
```diff
@@ -8,7 +8,7 @@ django-cached-authentication-middleware==0.2.1
 django-proxy==1.0.2
 django-redis-cache==1.6.5
 git+ssh://git@github.com/Harvard-University-iCommons/django-auth-lti.git@v1.2.5#egg=django-auth-lti==1.2.5
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.10.19#egg=django-icommons-common==1.10.19
+git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.13#egg=django-icommons-common==1.13
```
elliottyates: Note to make this work at the moment you'll have to pull https://github.com/Harvard-University-iCommons/django-icommons-common/pull/157 locally (or reference the branch from github), as the tag hasn't been created yet.

canvas_account_admin_tools/templates/canvas_account_admin_tools/_no_tools_available.html
```diff
@@ -0,0 +1,20 @@
+{# todo: fix styles, or move a customizable template into icommons_common #}
```
elliottyates: The styling on this works, and ultimately it should live in django-icommons-ui (typo in the code here) so it can be updated in just one place, like the default unauthorized message.

canvas_account_admin_tools/migrations/0002_school_permissions_20160826_1252.py
```diff
@@ -0,0 +1,43 @@
...
+# intialize permissions for all apps to the defaults (TLT-2650)
+PERMISSION_NAMES = settings.TOOL_PERMISSIONS.values()
+SCHOOL_PERMISSION_DATA = itertools.product(
+    PERMISSION_NAMES,
+    DEFAULT_SCHOOLS,
+    DEFAULT_ROLES)
+
+
+def create_school_permissions(apps, schema_editor):
+    SchoolPermission = apps.get_model('lti_permissions', 'SchoolPermission')
```
elliottyates: As suggested by @cmurtaugh, perhaps we can:
1. add account admin permissions (default)
2. figure out the additional permissions based on what's in the `LtiPermission` table and automatically generate those `SchoolPermission`s.

canvas_account_admin_tools/templates/canvas_account_admin_tools/_no_tools_available.html
```diff
@@ -0,0 +1,20 @@
...
+        <h2 class="ui-state-error">
+          <i class="fa fa-warning error-icon"></i>
+          Not available
+        </h2>
+        <div class="ui-error-msg">
+          <p>
+            You do not currently have access to any of the tools available in
+            Admin Console. If you think you should have access, please use the
+            ‘Help’ link in the bottom left corner of the screen to contact
+            Canvas support from Harvard.
```
elliottyates: Needs a closing `</p>`?

canvas_account_admin_tools/templates/canvas_account_admin_tools/_no_tools_available.html
elliottyates: Oops, I mean the styling _doesn't work_ at the moment -- will updated the original comment.

canvas_account_admin_tools/migrations/0002_school_permissions_20160826_1252.py
elliottyates: As discussed with @sapnamysore we're going to do the above 1 and 2 steps in the lti_permissions migrations now.

elliottyates: (so we can remove this migration here)

canvas_account_admin_tools/settings/base.py
```diff
@@ -263,9 +263,17 @@
...
 ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = False
 
-PERMISSION_ACCOUNT_ADMIN_TOOLS = 'account_admin_tools'
-PERMISSION_PEOPLE_TOOL = 'people_tool'
-PERMISSION_XLIST_TOOL = 'cross_listing'
+# TOOL_PERMISSIONS is used to map app names to SchoolPermission.permission names
+# to determine visibility of tools on dashboard and access to tool views/pages
+TOOL_PERMISSIONS = {
+    # NOTE: canvas_account_admin_tools permission name is shortened to
+    # account_admin_tools due to SchoolPermission.permission field max length
```
elliottyates: We could leave this dict as-is, and allow long key names to translate down to shorter ones for the sake of readable permission names in the SchoolPermission tables. The set of SchoolPermissions that need to be displayed to the user in TLT-2651 would be the `values()` of this dict. We could also simplify this to a `list` that just contains the values of this dict here, and have the `views` code use the same name that's in the list. Then the simplified `TOOL_PERMISSIONS` list can just be the keys. We lose the ability to make the dict keys be more meaningful, but a simple code comment can clarify a funky/terse/abbreviated key name, or we could even change the `SchoolPermission` model to accept permissions with names longer than 25 characters (not sure where that came from in the first place).

I lean towards a list here, with code comments to clarify any confusing permissions. Then referring to them from `lti_permissions` doesn't require the `values()` call (see e.g. https://github.com/Harvard-University-iCommons/django-icommons-common/pull/157/files#diff-1769bdf11052a9b2c8f7db98d3e9bdb4R12).

canvas_account_admin_tools/settings/base.py
```diff
@@ -263,9 +263,17 @@
...
+    'canvas_account_admin_tools': 'account_admin_tools',  # dashboard
+    'conclude_courses': 'conclude_courses',  # external tool
+    'course_info': 'course_info',
+    'cross_list_courses': 'cross_list_courses',
```
sapnamysore: Shouldn't this be 'cross_listing' as per https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/170/files#diff-6657219d5c6186d4b0ad46cf3bd746dfL268?

[For the dynamin query to find existing cross listing permissions?]

canvas_account_admin_tools/settings/base.py
sapnamysore: I am good with using list here and also increasing the column width  in the SchoolPermission table to be beyond 25. Now maybe a good time to do it since we are just creating it?

canvas_account_admin_tools/settings/base.py
elliottyates: Yeah, if we're going to have to be backwards-compatible, we will need to use the existing LtiPermission names, or at least map them somehow. This code was written assuming we could use new names, before our discussion about automatically mapping them. Good catch.

canvas_account_admin_tools/settings/base.py
elliottyates: Sounds good to me.

canvas_account_admin_tools/views.py
```diff
@@ -55,40 +55,27 @@ def lti_launch(request):
 
 
 @login_required
-@lti_role_required(const.ADMINISTRATOR)
-@lti_permission_required(settings.PERMISSION_ACCOUNT_ADMIN_TOOLS)
+@lti_permission_required(PERMISSIONS['canvas_account_admin_tools'])
```
sapnamysore: Hmm, thinking aloud. We would need the dict format to enable this right? If we change the TOOL_PERMISSIONS to a list, we loose the ability to selectively pick the appropriate permissions for the decorator here and for the different tools below?

elliottyates: Right, we would just have to use the permission name here. There would be no verification that it would exist in the TOOL_PERMISSIONS list; we'd just have to make sure they matched up e.g. by eye.

elliottyates: (so here it would change to:)
```
@lti_permission_required('account_admin_tools')
```

sapnamysore:
👍

cross_list_courses/views.py
```diff
@@ -4,23 +4,21 @@
...
-from django_auth_lti.decorators import lti_role_required
 
 logger = logging.getLogger(__name__)
+PERMISSIONS = settings.TOOL_PERMISSIONS
 
 
 @login_required
-@lti_role_required(const.ADMINISTRATOR)
-@lti_permission_required(settings.PERMISSION_XLIST_TOOL)
+@lti_permission_required(PERMISSIONS['cross_list_courses'])
```
sapnamysore: reminder to ourselves to validate if this  value  will be 'cross_list_courses' or 'cross_listing' for backward compatibility

cross_list_courses/views.py
```diff
@@ -4,23 +4,21 @@
...
 @require_http_methods(['GET'])
 def index(request):
     return render(request, 'cross_list_courses/index.html')
 
 @login_required
-@lti_role_required(const.ADMINISTRATOR)
-@lti_permission_required(settings.PERMISSION_XLIST_TOOL)
+@lti_permission_required(PERMISSIONS['cross_list_courses'])
```
sapnamysore: same as above

canvas_account_admin_tools/requirements/base.txt
```diff
@@ -8,8 +8,8 @@ django-cached-authentication-middleware==0.2.1
 django-proxy==1.0.2
 django-redis-cache==1.6.5
 git+ssh://git@github.com/Harvard-University-iCommons/django-auth-lti.git@v1.2.5#egg=django-auth-lti==1.2.5
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.10.19#egg=django-icommons-common==1.10.19
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-ui.git@v1.2.0#egg=django-icommons-ui==1.2.0
+git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.13#egg=django-icommons-common==1.13
+git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-ui.git@v1.2.1#egg=django-icommons-ui==1.2.1
```
elliottyates: Note to make this work at the moment you'll have to pull Harvard-University-iCommons/django-icommons-common#157 and https://github.com/Harvard-University-iCommons/django-icommons-ui/pull/82 locally (or reference the branches from github), as the tags haven't been created yet.

canvas_account_admin_tools/settings/base.py
```diff
@@ -263,9 +263,17 @@
 ICOMMONS_REST_API_TOKEN = SECURE_SETTINGS.get('icommons_rest_api_token')
 ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = False
 
-PERMISSION_ACCOUNT_ADMIN_TOOLS = 'account_admin_tools'
-PERMISSION_PEOPLE_TOOL = 'people_tool'
-PERMISSION_XLIST_TOOL = 'cross_listing'
+# TOOL_PERMISSIONS is used to map app names to SchoolPermission.permission names
```
elliottyates: This is no longer a map, so comment needs to be updated.

canvas_account_admin_tools/views.py
```diff
@@ -55,40 +56,38 @@ def lti_launch(request):
...
-@lti_permission_required(settings.PERMISSION_ACCOUNT_ADMIN_TOOLS)
+@lti_permission_required('account_admin_tools')
 @require_http_methods(['GET'])
 def dashboard_account(request):
-    custom_canvas_account_sis_id = request.LTI['custom_canvas_account_sis_id']
-    custom_canvas_membership_roles = request.LTI['custom_canvas_membership_roles']
-    conclude_courses = settings.CONCLUDE_COURSES_URL
+    tools = ['conclude_courses',
+             'course_info',
+             'cross_listing',
```
elliottyates: Not actually a tool name, but a permission name; perhaps changing this list variable's name, or making a comment that it needs to match the item in `TOOL_PERMISSIONS`, would make this clearer.

elliottyates:
(Closing for now. This will require a fresh set of eyes when we get around to bringing the new permission scheme to caat)

PR #173, opened 2016-11-03
course info - selenium tests for editing SB/ILE course info

elliottyates:
- reintroduces tests that were originally written for TLT-2376
- course info page views are implemented with angular, so we need an angular-aware way to wait for the page to become stable before we attempt to access angular-managed DOM elements

### DIFF
--- a/selenium_tests/course_info/page_objects/course_info_detail_page_object.py
+++ b/selenium_tests/course_info/page_objects/course_info_detail_page_object.py
@@ -1,7 +1,8 @@
-from selenium.common.exceptions import NoSuchElementException
-from selenium.common.exceptions import TimeoutException
+from selenium.common.exceptions import NoSuchElementException, TimeoutException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.wait import WebDriverWait
+from selenium.webdriver.support.expected_conditions \
+    import element_to_be_clickable
 
 from selenium_tests.course_info.page_objects.course_info_base_page_object \
     import CourseInfoBasePageObject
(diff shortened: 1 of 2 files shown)


selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -77,8 +82,20 @@ def submit_was_successful(self):
...
-        # server (which in this case means the spinner disappears and the input
-        # element becomes visible)
-        WebDriverWait(self._driver, 10).until(lambda s: s.find_element(
-            *Locators.INPUT_BY_FIELD_NAME(field_name)).is_displayed())
+        self._wait_for_angular()
+        # element_to_be_clickable works better with Angular than is_displayed()
+        WebDriverWait(self._driver, 10).until(element_to_be_clickable(
+            Locators.INPUT_BY_FIELD_NAME(field_name)))
+
+    def _wait_for_angular(self):
```
elliottyates: This is something we're eventually going to want in selenium_common. (I think we could move a configurable `WebDriverWait` in there too, for brevity and configurability's sake.) Punting for now until we're sure it's a stable way to do this. :)

hktest:
Ran regression_suite on the project both locally and non-locally and they all passed with the fix.  The "wait_for_angular" method seems to work, so :+1:  (for me)  
regression_suite_course_info

selenium_tests/course_info/page_objects/course_info_detail_page_object.py
```diff
@@ -77,8 +82,20 @@ def submit_was_successful(self):
         return True
 
     def _wait_for_input_field_to_be_visible(self, field_name):
-        # only want to access input elements once the values are loaded from the
-        # server (which in this case means the spinner disappears and the input
-        # element becomes visible)
-        WebDriverWait(self._driver, 10).until(lambda s: s.find_element(
-            *Locators.INPUT_BY_FIELD_NAME(field_name)).is_displayed())
+        self._wait_for_angular()
+        # element_to_be_clickable works better with Angular than is_displayed()
```
elliottyates: Strange -- this is not showing up in our slack `#github` channel... just noticed it because of the PR-level (vs file-level) comment that did show up in slack.

sapnamysore:
Looks good. Tested and verified that things work.. 👍 

I spent some time trying to isolate the issue as I was looking at all of them as new code(in spite of your comment :-)). It was much easier once I isolated the delta..I would suggest the for similar changes in future, we should have  them in 2 commits - one reintroducing the old code and another with the fix.

PR #174, opened 2016-11-03
Added course section info and modified labels to be consistent

sapnamysore:
https://jira.huit.harvard.edu/browse/TLT-2739


elliottyates:
👍 looks good to me.

It would be great to get these into shared components. I think we have some tech debt coming up (e.g. [TLT-2267](https://jira.huit.harvard.edu/browse/TLT-2267)) that will give us a start, and then we can build from there.

PR #176, opened 2016-11-22
Tlt 2645 selenium tests for bulk create

hktest:
- Refactored selenium tests with the move over from icommons_lti_tool to canvas_account_admin (particularly around using admin dashboard instead of passing into direct bulk create URL) 
- Fixed test data 
- Added explicit wait to better selenium performance

Tested, and ran regression test for the project successfully.

### DIFF
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -43,16 +43,16 @@ SELENIUM_CONFIG = {
             'course_short_title': 'Selenium Auto Test 101',
             'course_title': 'Selenium Automated Test Course 101',
             'template': 'None',  # No Template
-            'term': '1',
+            'term': 'string:5209',
 
             #TLT-2522 - Testing course with and without registrar_code_display
             'course_with_registrar_code_display': {
-                'registrar_code_display': 'selenium_test',
-                'sis_id_value': '341697',
+                'registrar_code_display': 'Automated_Test',
+                'sis_id_value': '362568',
             },
             'course_without_registrar_code_display': {
-                'registrar_code_display': '109581',
-                'sis_id_value': '342916',
+                'registrar_code_display': '117138',
+                'sis_id_value': '360031',
             },
         },
     },
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -54,11 +54,7 @@ SELENIUM_CONFIG = {
                 'registrar_code_display': '109581',
                 'sis_id_value': '342916',
             },
-
-
         },
-        'url': 'accounts/10/external_tools/11',
-
     },
    'canvas_base_url': CANVAS_URL,
    'course_info_tool': {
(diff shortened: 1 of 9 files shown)


canvas_account_admin_tools/settings/local.py
```diff
@@ -43,22 +43,18 @@
             'course_short_title': 'Selenium Auto Test 101',
             'course_title': 'Selenium Automated Test Course 101',
             'template': 'None',  # No Template
-            'term': '1',
+            'term': 'string:5209',
```
sapnamysore: Talked offline about the change in term: That's because term=1 doesn't exist anymore

selenium_tests/bulk_create/create_new_ci_tests.py
```diff
@@ -54,11 +54,10 @@ def test_create_site_without_short_title(self):
             self.test_data['course_code'],
             self.test_data['course_title'],
             '',  # simulates no short title entered in form
-            self.test_data['term'],
             expected_course_code_text)
 
     def _test_create_site(self, course_code, course_title, course_short_title,
-                          course_term, expected_course_code):
```
sapnamysore: Do we no longer need the term for these tests?

sapnamysore:
A couple of clarifications, otherwise looks good 👍 .

We can co ordinate when you are ready to merge to my branch and retest with the latest changes.

canvas_account_admin_tools/settings/local.py
hktest: Indeed, tests were failing locally due to term 1 not found.  Updated to fix tests.  
term_existing_course

selenium_tests/bulk_create/create_new_ci_tests.py
hktest: Great point/call!   I initially removed it because it wasn't critical that a term had to be specified for new course creation (it would default to the first one in the dropdown). Also, we were running in "term 1  not found" errors. However, per our discussions, I investigated on making the terms work for these tests.  It turns out there is a  consistency issue in the code.  In the "create new site for existing course" term dropdown, the option value is "string:xxxx", while In the create new site for new course" term dropdown, the option value is "numbers:xxxx".   I've updated the test data in local.py for it can fetch the right term value, updated the code to include terms again, and verified that it works after these changes.

selenium_tests/bulk_create/bulk_create_base_test_case.py
```diff
@@ -2,35 +2,36 @@
...
-        else:
-            print '(XID user {} already logged in)'.format(cls.username)
+class BulkCreateBaseTestCase(AccountAdminBaseTestCase):
+
+    def setUp(self):
+
+        super(AccountAdminBaseTestCase, self).setUp()
+
+        self.test_data = settings.SELENIUM_CONFIG['bulk_create']['test_data']
+        self.test_data_course1 = settings.SELENIUM_CONFIG['bulk_create'][
```
hktest: Done, and renamed.  Thanks!

PR #178, opened 2016-12-01
tlt-2815: site creator - create new course - registrar code validation

elliottyates:
- check for illegal characters, including blanks, in registrar_code and warn when input contains one or more


elliottyates:
This code was reviewed and tested in https://github.com/Harvard-University-iCommons/icommons_lti_tools/pull/461, and has been tested in this project as well, so going ahead with merge.

PR #180, opened 2016-12-01
Task/sapnamysore/tlt 2645/fix breadcrumb style

sapnamysore:
Removed the custom lti-css/bulkSiteCreator.css that was in place that had partial styles from bootstrap and then some custom ones. Swapped it with standard bootstrap.css in all screens. Moved from styles from bulkSiteCreator.css to app.css to fix some positioning styles.

The screens in Canvas Site Creator are now more consistent with the rest of teh tools in admin console. Any custom styles can be fixed or added as enhancement once @Vittorio2015 and the PO take a look.

The code is deployed on Dev https://canvas.dev.tlt.harvard.edu/accounts/8/external_tools/44

### DIFF
--- a/canvas_site_creator/templates/canvas_site_creator/base.html
+++ b/canvas_site_creator/templates/canvas_site_creator/base.html
@@ -8,16 +8,11 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">
     <link rel="stylesheet" type="text/css" href="{% static 'font-awesome-4.5.0/css/font-awesome.min.css' %}">
     <link rel="stylesheet" href="{% static 'canvas_site_creator/css/app.css' %}">
-    <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.min.css' %}" media="screen"/>
 
-    {% block css %}
-      {% if debug %}
-        <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.css' %}" media="screen"/>
-      {% else %}
-        <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.min.css' %}" media="screen"/>
-      {% endif %}
-    {% endblock css %}
 {% if debug %}
+    {#  Note: retain the order for proper styles, first load the jquery.datatables.css and then bootstrap   #}
+    <link rel="stylesheet" type="text/css" href="{% static 'datatables-1.10.7/media/css/jquery.dataTables.css' %}" media="screen"/>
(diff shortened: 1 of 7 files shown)


elliottyates:
Hmm, I would prefer to approach this by standardizing the styles in this app with the ones in the other apps -- e.g. use bootstrap.css in all pages (it's currently not included on these two), and removing the `lti-css/bulkSiteCreator.css` in `base.html`, so there aren't conflicts. It would be very hard otherwise to make everything work (for instance, at the moment there are other subtle differences between the headers in these two pages and the other three in this app -- like the link underlining, explicit background coloring, etc). Switching to a common set of styles and removing the extraneous overrides (much of the `lti-css/bulkSiteCreator.css` is following our previous customized bootstrap stuff, and can be swapped out safely for the regular packaged bootstrap CSS) will mean our headers, datatable formats, etc will be consistent across all apps in this project as well.

While we're at it, in general these CSS/JS blocks can all be cleaned up, it looks like. Each page has implemented its additions to base.html a little differently. I think making them consistent and cutting out the `lti-css/bulkSiteCreator.css` does mean a few things (like padding) will change, but we can fix that up much easier by editing `app.css` at that point and adding useful classes to target things so they don't conflict with shared CSS.

I'm happy to take a first pass at that. If we want to proceed with tweaking these two pages, let's try and get most of the computed CSS properties of the header text to be (a) consistent across all five screens in the app, and (b) consistent with other apps in the project.

sapnamysore:
Sure, good with switching out the custom bootstrap to the standard one. The introduction of the standard bootstrap messed up pretty much everything in the 2 screens..

canvas_site_creator/templates/canvas_site_creator/base.html
```diff
@@ -7,9 +7,16 @@
...
+    <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.min.css' %}" media="screen"/>
+
+    {% block css %}
+      {% if debug %}
+        <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.css' %}" media="screen"/>
+      {% else %}
+        <link rel="stylesheet" type="text/css" href="{% static 'bootstrap-3.3.6/dist/css/bootstrap.min.css' %}" media="screen"/>
+      {% endif %}
+    {% endblock css %}
 {% if debug %}
```
elliottyates: As per the issue Chrissy reported, if we get the bootstrap JS into this block, and remove the tooltip.js and tooltip.css stuff from the course selection page, that appears to fix the weird modal behavior.

elliottyates:
One breaking change that PO found, and some suggestions. Probably worth another quick functionality test after that -- and maybe just cleaning up the whole CSS/JS on each of the five pages in general (taking out custom JS/CSS and consolidating standard stuff in base.html).

elliottyates:
Looks great! A few suggestions for cleanup but so nice to see this in better shape than it was in icommons_lti_tools.

PR #181, opened 2016-12-07
Hchow/bulk site creator term update

hktest:
- Looking to merge a commit that was previously thought merged.  This is cleanup of the variable term to lookup by term text, rather than term id (which had been changed with the addition of a prefix (i.e. "numbers", "string" added to the term ID)  
- Per feedback from Sapna, also updated the "bulk_create" references at the file/code level with the tool name change, which is now canvas_site_creator, for consistency.  
- Ran regression_suite.py for the project after these changes.  Looks ok.  
selenium_tests_canvas_account_admin
@sapnamysore


sapnamysore:
👍 Looks good!
Glad to see the renaming of the selenium tests folder to match the folder name of the app in the rest of the code.

PR #182, opened 2016-12-07
Added the newly created librarian role (role_id=16), to the list cont…

sapnamysore:
…rolling the dropdown roles
Note: The role has been added in Canvas as well as the DB in both prod and QA. The role_id(16) in user_role table needs to match in both environments as it is set in the base.py. The canvas_role_id can be different

This is deployed on dev:
https://canvas.dev.tlt.harvard.edu/accounts/10/external_tools/79


elliottyates:
👍 looks good

PR #183, opened 2016-12-12
tlt-2645: site creator - cron jobs 

elliottyates:
- cron jobs for process_async_jobs and finalize_bulk_create_jobs (moved from icommons_ext_tools)


bermudezjd:
You actually don't need to specify `settings` since manage.py defaults to using aws, but I'll leave it up to you whether you want to enforce it or not.

PR #186, opened 2017-01-03
tlt-1805: publish courses - term dropdown tweaks

elliottyates:
Term dropdown only shows active terms that are not yet concluded.

Note that once this is merged we'll need to update QA data to be able to see active terms. :)

@sapnamysore @hktest

### DIFF
--- a/publish_courses/static/publish_courses/js/controllers/IndexController.js
+++ b/publish_courses/static/publish_courses/js/controllers/IndexController.js
@@ -10,6 +10,7 @@
         //   {alertType: 'bootstrap alert class', text: 'actual message'}
         $scope.message = null;
         $scope.operationInProgress = false;
+        $scope.currentDate = new Date();
         $scope.school = {id: AppConfig.school};
         $scope.selectedTerm = null;
         $scope.terms = null;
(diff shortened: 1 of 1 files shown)


publish_courses/static/publish_courses/js/controllers/IndexController.js
```diff
@@ -83,5 +89,7 @@
             $scope.clearMessages();
             $scope.loadCoursesSummary();
         };
+
+        $scope.initialize();
```
sapnamysore: Was the initialize happening  earlier with the ng-init?

sapnamysore:
Looks good 👍

publish_courses/static/publish_courses/js/controllers/IndexController.js
elliottyates: No, it was just happening in the main body of the controller (when the controller was instantiated). By moving it into an initialize() function, the order of the `$scope` function declarations doesn't matter anymore (otherwise any functions called by the initialization code would need to be defined above that code in the main body of the controller function). (Incidentally, `ng-init` is [not recommended for kicking off controller code](https://docs.angularjs.org/api/ng/directive/ngInit), apparently -- I was interested in that too, because I thought then jasmine tests would be a little easier.)

elliottyates:
Thanks!

PR #191, opened 2017-01-06
Hchow/tlt 1805 bulk publish selenium tests with 1805 code

hktest:
- Set up selenium PO structure for the Publish Courses Tool 
- Added in simple tests to test workflow 
- This branch was rebased with canvas_account_admin_tools/tree/task/sapnamysore/tlt-1805/publish-courses.

### DIFF
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -112,7 +112,7 @@ SELENIUM_CONFIG = {
             'account': 10,
             #'courses': [],
             'term': "sis_term_id:2016-2",
-            'dry_run': True
+            'dry_run': False
             },
      },
 
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -105,8 +105,8 @@ SELENIUM_CONFIG = {
             'relative_url': 'courses/2601'
         },
         'test_terms': {
-            'term_with_unpublished_courses': 'Summer 2017',
-            'term_with_all_published_courses': 'Full Year 2016',
+            'with_unpublished_courses': 'Summer 2017',
+            'with_all_published_courses': 'Full Year 2016',
         },
         # When testing on a term with large number of courses, test in dry
         # run mode first to verify that numbers match expected results.
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -108,11 +108,11 @@ SELENIUM_CONFIG = {
         # When testing on a term with large number of courses, test in dry
         # run mode first to verify that numbers match expected results.
         'op_config': {
-            'published': 'true',
             'account': 10,
+            'published': 'false',
             #'courses': [],
             'term': "sis_term_id:2016-2",
-            'dry_run': False
+            # 'dry_run': False
             },
      },
 
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -102,7 +102,7 @@ SELENIUM_CONFIG = {
 
     'publish_courses': {
         'test_terms': {
-            'term_with_unpublished_courses': 'Spring 2016-2017',
+            'term_with_unpublished_courses': 'Fall 2016',
             'term_with_all_published_courses': 'Full Year 2016',
         },
         # When testing on a term with large number of courses, test in dry
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -36,7 +36,7 @@ SELENIUM_CONFIG = {
       'relative_url': 'accounts/10/external_tools/79',  # dev (Admin Console)
       #'relative_url': 'accounts/10/external_tools/99',  # qa (Admin Console)
    },
-   'canvas_site_creator': {
+    'canvas_site_creator': {
         'test_data': {
             'course_code': 'Selenium-Automated',  # defaults to ILE
             'course_group': 'Anthropology',
@@ -99,11 +99,22 @@ SELENIUM_CONFIG = {
    'icommons_rest_api': {
       'base_path': 'api/course/v2'
    },
-    # TODO: Need to set up proper terms for use with the test cases
+
     'publish_courses': {
-        'term_with_unpublished_courses': 'Fall 2016',
-        'term_with_all_published_courses': 'Fall 2016',
-    },
+        'test_terms': {
+            'term_with_unpublished_courses': 'Spring 2016-2017',
+            'term_with_all_published_courses': 'Full Year 2016',
+        },
+        # When testing on a term with large number of courses, test in dry
+        # run mode first to verify that numbers match expected results.
+        'op_config': {
+            'published': 'true',
+            'account': 10,
+            #'courses': [],
+            'term': "sis_term_id:2016-2",
+            'dry_run': True
+            },
+     },
 
    'run_locally': SECURE_SETTINGS.get('selenium_run_locally', False),
    'selenium_username': SECURE_SETTINGS.get('selenium_user'),
--- a/canvas_account_admin_tools/settings/local.py
+++ b/canvas_account_admin_tools/settings/local.py
@@ -101,8 +101,11 @@ SELENIUM_CONFIG = {
    },
 
     'publish_courses': {
+        'test_course': {
+            'relative_url': 'courses/2601'
+        },
         'test_terms': {
-            'term_with_unpublished_courses': 'Fall 2016',
+            'term_with_unpublished_courses': 'Summer 2017',
             'term_with_all_published_courses': 'Full Year 2016',
         },
         # When testing on a term with large number of courses, test in dry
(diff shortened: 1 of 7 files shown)


canvas_account_admin_tools/settings/local.py
```diff
@@ -99,6 +99,26 @@
    'icommons_rest_api': {
       'base_path': 'api/course/v2'
    },
+
+    'publish_courses': {
+        'test_course': {
+            'relative_url': 'courses/2601'
+        },
+        'test_terms': {
+            'term_with_unpublished_courses': 'Summer 2017',
```
elliottyates: suggestion for brevity -- these are subkeys of `test_terms`, so the `term_` bit of this key could probably be dropped without losing readability in the code.

selenium_tests/publish_courses/publish_courses_tests.py
```diff
@@ -0,0 +1,92 @@
...
+        This is a setup and cleanup method to unpublish courses in a term
+        """
+        # Get term information from Settings
+        op_config = self.term['op_config']
+
+        # Bulk update courses based on parameter passed in op_config
+        op = BulkCourseSettingsOperation(op_config)
+        op.execute()
+
+    def refresh_page(self, term):
```
elliottyates: Another thing you could do, which is faster, is re-select the term in the dropdown. That will refresh the summary. So you could create a `refresh_course_summary()`, which would re-select the currently-selected item, or just use your PO's existing `select_term()` a second time.

selenium_tests/publish_courses/publish_courses_base_test_case.py
```diff
@@ -0,0 +1,39 @@
...
+
+        #  Test course URL
+        self.base_url = settings.SELENIUM_CONFIG['canvas_base_url']
+        self.test_course_relative_url = settings.SELENIUM_CONFIG[
+            'publish_courses']['test_course']['relative_url']
+        self.test_course_url = urljoin(self.base_url,
+                                    self.test_course_relative_url)
+
+        # Test data for use in tests
+        self.term = settings.SELENIUM_CONFIG['publish_courses']
```
elliottyates: Why is this called `self.term`? I find that a bit confusing. Maybe `self.config` or `self.test_settings`?

selenium_tests/publish_courses/publish_courses_base_test_case.py
```diff
@@ -0,0 +1,39 @@
...
+            self.acct_admin_dashboard_page.get(self.TOOL_URL)
+
+        # navigate to Publish Courses tool
+        self.acct_admin_dashboard_page.select_publish_courses_link()
+
+        # check if page is loaded (which will also set the focus on the tool)
+        self.assertTrue(self.main_page.is_loaded())
+
+        #  Focus on tool frame
+        self.main_page.focus_on_tool_frame()
```
elliottyates: Should already have focus, if you were able to click on the dashboard link, but this doesn't hurt.

selenium_tests/publish_courses/publish_courses_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+        main_page = MainPageObject(self.driver)
+
+        #  Select a term from the term dropdown
+        term_with_unpublished_courses = self.term['test_terms'][
+            'term_with_unpublished_courses']
+        main_page.select_term(term_with_unpublished_courses)
+
+        # If the Publish Courses button is enabled, click on it.
+        if main_page.is_publish_all_button_enabled():
```
elliottyates: This can be simplified as:
```
if not main_page.is_publish_all_button_enabled():
    self.bulk_unpublish_canvas_sites()
    ...

# main code path (either way you want to publish)
main_page.publish_courses()
```

selenium_tests/publish_courses/publish_courses_tests.py
```diff
@@ -0,0 +1,92 @@
...
+            main_page.publish_courses()
+        else:
+            #  If "Publish all" button is not enabled, unpublish courses
+            #  first to ensure that the publish functionality can be tested.
+            self.bulk_unpublish_canvas_sites()
+
+            #  Refresh the page to see the changes
+            self.refresh_page(term_with_unpublished_courses)
```
elliottyates: See note about re-selecting term instead of refresh_page.

Also note that it may take a few seconds for all terms to be processed, but probably by the time you requery (even with the quicker re-select term method) at least one term will be unpublished, which will enable the button).

selenium_tests/publish_courses/publish_courses_tests.py
```diff
@@ -0,0 +1,92 @@
...
+            #  Verify that the Publish Button is clickable
+            self.assertTrue(main_page.is_publish_all_button_enabled())
+            #  Click on "Publish all" button
+            main_page.publish_courses()
+
+        #  Verify that a confirmation appears on screen
+        self.assertTrue(main_page.result_message_visible())
+
+        #  Go to a test course to spotcheck
+        self.driver.get(self.test_course_url)
```
elliottyates: This will work if there are very few courses in the term... since it's asynchronous, there's no guarantee that this course will be done by the time we navigate there. In practice though, if there are very few courses, it probably will be! So if it's working consistently, that's good -- just raising it so we're aware of the potential for this to not work as hoped.

selenium_tests/publish_courses/publish_courses_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+        main_page = MainPageObject(self.driver)
+
+        #  Select a term from the term dropdown
+        term_with_all_published_courses = self.term['test_terms'][
+            'term_with_all_published_courses']
+        main_page.select_term(term_with_all_published_courses)
+
+        # If there are courses not yet published, publish it now.
+        if main_page.is_publish_all_button_enabled():
```
elliottyates: This can be simplified as:

```
 if main_page.is_publish_all_button_enabled():
    # publish, etc...

# main code path (want to make this assertion either way)
self.assertFalse(main_page.is_publish_all_button_enabled())
```

selenium_tests/publish_courses/page_objects/main_page.py
```diff
@@ -0,0 +1,62 @@
...
+    def publish_courses(self):
+        publish_courses = self.find_element(*Locators.PUBLISH_ALL_BUTTON)
+        publish_courses.click()
+
+    def result_message_visible(self):
+        """
+        Verify that the result message apepears after clicking on Publish All
+        """
+        try:
+            WebDriverWait(self._driver, 60).until(
```
elliottyates: Probably don't need 60sec here

selenium_tests/publish_courses/page_objects/main_page.py
```diff
@@ -0,0 +1,62 @@
...
+
+class Locators(object):
+    PUBLISH_ALL_BUTTON = (By.ID, "publishAllButton")
+    TERM_DROPDOWN = (By.ID, "dropdownMenuTerm")
+    RESULT_CONFIRMATION = (By.ID, "result-message")
+
+    @classmethod
+    def FIND_ELEMENT_WITH_TEXT(cls, text_value):
+        """Locates the xpath, given the text value"""
+        return By.XPATH, "//a[contains(.,'{}')]".format(text_value)
```
elliottyates: This finds anchor elements, no? so maybe `FIND_ELEMENT_WITH_TEXT` could be `FIND_LINK_WITH_TEXT` or something?

elliottyates:
👍

canvas_account_admin_tools/settings/local.py
hktest: Good point, fixed.

selenium_tests/publish_courses/publish_courses_tests.py
hktest: Nice, did not occur to me to simplify it this way.  Fixed and thanks.

selenium_tests/publish_courses/publish_courses_tests.py
hktest: I intentionally set up a test term with only a few courses  This check should hold and so far is working consistently.  For clarity purposes, I will add in a comment to the code to explicitly note this possibility.

selenium_tests/publish_courses/page_objects/main_page.py
hktest: Fixed.

selenium_tests/publish_courses/page_objects/main_page.py
hktest: Indeed.  More descriptive, and fixed!

selenium_tests/publish_courses/publish_courses_base_test_case.py
hktest: Tested and you're right - it's got focus already.  Removing this line.

selenium_tests/publish_courses/publish_courses_base_test_case.py
hktest: Hmm. Good catch - thanks!

selenium_tests/publish_courses/publish_courses_tests.py
hktest: At the time of building the tests, I didn't realize that reselecting the term refreshes the status.  This is the more accurate way to do it and I'll clean it up to re-select term.  Thanks!

selenium_tests/publish_courses/publish_courses_tests.py
hktest: See note below.  I will fix this one, thanks! :)

elliottyates:
(testing - should show in #testchannel)

PR #196, opened 2017-01-17
Added some tests for api.py. 

sapnamysore:
Note: SummaryList test case is skipped as it's not complete. But wanted to push the ones that are done so it can be merged if possible.

### DIFF
--- a/publish_courses/tests/test_api.py
+++ b/publish_courses/tests/test_api.py
@@ -2,15 +2,16 @@ import json
 from mock import patch, Mock
 import unittest
 
-from django.test import TestCase, RequestFactory
-import publish_courses.api as api
+from rest_framework.test import APITestCase
 from rest_framework.exceptions import PermissionDenied
 
+import publish_courses.api as api
 
-class ApiTestCase(TestCase):
+
+class PublishCoursesAPITestCase(APITestCase):
 
     def setUp(self):
-        super(ApiTestCase, self).setUp()
+        super(PublishCoursesAPITestCase, self).setUp()
 
         self.post_data = {
             'account': 'sis_account_id:school:abc',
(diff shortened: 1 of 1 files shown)


publish_courses/tests/test_api.py
```diff
@@ -1,11 +1,111 @@
-# Test TODOs
-# BulkPublishListCreate:
-# invalid account raises permission denied
-# invalid LTI session raises permission denied
```
elliottyates: looks like we lost this one in the mix?

publish_courses/tests/test_api.py
```diff
@@ -1,11 +1,111 @@
...
-
-# SummaryList:
-# invalid LTI session raises permission denied
-# returns data as expected on successful enqueue()
-# exception raises 500
+import json
+from mock import patch, Mock
+import unittest
+
+from django.test import TestCase, RequestFactory
```
elliottyates: We should probably use the DRF's test case class instead

publish_courses/tests/test_api.py
```diff
@@ -1,11 +1,111 @@
...
+        }
+
+        self.request = RequestFactory().post(self.post_url, **request_kwargs)
+        # these values don't matter; the keys need to be present
+        # for the authorization decorator patches to work
+        self.request.LTI = self.lti_data
+        self.request.data = data
+        self.request.user = self.user
+        bulk_publish = api.BulkPublishListCreate()
+        bulk_publish.request = self.request
```
elliottyates: initializing the view with arguments `request=self.request` will do the same thing here.

publish_courses/tests/test_api.py
```diff
@@ -1,11 +1,111 @@
...
+    def test_bulk_publish_list_create_missing_data(self):
+        """
+        Invalid/missing data raises Exception
+        """
+        data = {}
+        with self.assertRaises(Exception):
+            self._prep_request_and_post(data)
+
+    # todo
+    # exception raises 500
```
elliottyates: we could duplicate `test_bulk_publish_list_create_missing_data()` with a live server, but I think it's fine to trust that DRF will return a 500 if we raise a standard `Exception`, as it's supposed to.

PR #193, opened 2017-01-10
Task/sapnamysore/tlt 1805/publish courses

sapnamysore:
This PR contains changes for the [tlt-1805](https://jira.huit.harvard.edu/browse/TLT-1805) : a new tool to Bulk publish Canvas sites.  It includes the front end, back end and bulk utilities. The unit tests and Jasmine tests are being worked on in parallel, and shouldn't hold up this PR . 

It depends on this PR :https://github.com/Harvard-University-iCommons/django-icommons-common/pull/169 
[Note : once the above PR is merged, it's reference in  base.text will need to be fixed]

PR #190, opened 2017-01-05
Made the queue name configurable so it  can be reused by the ansible …

sapnamysore:
…scripts to deploy the rqworker in the various environments.
 Note: add   the following to secure.py:  
**rqworker_queue_name: 'bulk_publish_canvas_sites'**

This PR relates to https://github.com/Harvard-University-iCommons/tlt-ops/pull/238

PR #198, opened 2017-01-23
TLT-1805 publish courses - unit tests

elliottyates:
unit tests for async_operations and api
- publish_courses.api.BulkPublishListCreate now puts jobs on the queue defined in project settings (instead of hardcoded)
- more robust LTI session error handling for create job

### DIFF
--- a/publish_courses/api.py
+++ b/publish_courses/api.py
@@ -34,6 +34,8 @@ class ProcessSerializer(ModelSerializer):
 
 
 class LTIPermission(BasePermission):
+    message = 'Invalid LTI session.'
+
     def has_permission(self, request, view):
         return lti_permission_required_check(request, PC_PERMISSION)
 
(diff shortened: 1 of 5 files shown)


sapnamysore:
Nice work!  👍 for adding the appropriate links/documentation (eg: SimpleWorker, etc)

publish_courses/tests/test_api.py
```diff
@@ -1,125 +1,199 @@
...
         """
         Invalid/missing data raises Exception (REST framework converts that
-        into a status code 500)
+        into a response with status code 400)
         """
         data = {}
-        with self.assertRaises(Exception):
+        msg = 'account and term are required'
+        with self.assertRaisesRegexp(DRFValidationError, msg):
             self._prep_request_and_post(data)
```
elliottyates: Yes, good call, it also guards against future updates to the method signature that might add positional args.

publish_courses/api.py
```diff
@@ -101,7 +109,7 @@ def create(self, request, *args, **kwargs):
 
         process = Process.enqueue(
             bulk_publish_canvas_sites,
-            'bulk_publish_canvas_sites',
+            settings.RQWORKER_QUEUE_NAME,
```
elliottyates: Yeah I think we missed it here in this one spot.
(See #190 and #193)

PR #203, opened 2017-02-22
tlt-2900: cross list courses - fix to show user-friendly errors when …

elliottyates:
…deleting xlist_maps.

When the REST API DELETE view throws a ValidationError, it sends a simple list of errors encountered (currently, just one is possible). This is different from the Django rest framework's handling of serializer errors in the POST view, so the logic around where to look in the response is a bit different.

Basically, we transform the specific, but brief, API error into a user-friendly version if we encounter it; any other error(s) are reflected to the user as a generic failure message.

### DIFF
--- a/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
+++ b/cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
@@ -115,15 +115,29 @@
                     };
                 }, function DeleteFailed(response) {
                     $scope.handleAjaxErrorResponse(response);
-                    var errorMessage = (((response||{}).data||{}).detail||'');
-                    if (errorMessage.indexOf('has multiple site maps') > -1) {
-                         errorText =  'These courses cannot be de-cross-listed because '+
-                              primary + ' and/or '+secondary+' is currently associated with more than '+
-                             'one course site. Please contact '+
-                             'academictechnology@harvard.edu for further assistance.';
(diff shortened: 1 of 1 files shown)


cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
```diff
@@ -115,15 +115,29 @@
...
-                    if (errorMessage.indexOf('has multiple site maps') > -1) {
-                         errorText =  'These courses cannot be de-cross-listed because '+
-                              primary + ' and/or '+secondary+' is currently associated with more than '+
-                             'one course site. Please contact '+
-                             'academictechnology@harvard.edu for further assistance.';
-                    } else{
-                        errorText = 'Could not de-cross-list ' + primary +
-                            ' and ' + secondary + '. Please try again later.';
+                    var defaultErrorText = primary + ' could not be ' +
+                        'crosslisted with ' + secondary + ' at this ' +
```
sapnamysore: This should read 'could not be de-crosslisted..'

sapnamysore:
👍

cross_list_courses/static/cross_list_courses/js/controllers/ListController.js
elliottyates: Haha now how did that sneak in there! Thanks, good catch,

PR #206, opened 2017-06-12
Fix json serialization error by using simplejson

Chris-Thornton-Harvard:
The simplejson library is a well maintained encoder/decoder and has performance benefits over using the json library.


sapnamysore:
Looks good 👍 .  Could you verify and spot check that the other screens in caat render ok (or if they need this fix as well)

Chris-Thornton-Harvard:
Other screens are showing no similar errors. I will look into simplejson's implementation throughout the project.

PR #208, opened 2017-06-13
Create all button fix for account types other than course groups

Chris-Thornton-Harvard:
The bulk creation would only work when a course group id was passed in to the get_course_instance_query_set function.

This adds a check for course group, department and school from the passed in filters.
Also adds a filter for course instances with canvas course id's.

[TLT-2927](https://jira.huit.harvard.edu/browse/TLT-2927)


Chris-Thornton-Harvard:
This branch is currently deployed on dev for testing.

PR #210, opened 2017-06-23
Check if a CI’s subtitle and title are not the same

Chris-Thornton-Harvard:
Some schools set the title and subtitle to be the same content.
This creates a check before appending the CI subtitle to the title in the 'Search Courses' app.

Example courses to test this can be viewed on dev by using the 'Search Courses' app and searching in HKS and the title “ile”.
Course with a title and subtitle, not equal to each other: 378076
Course with a title and subtitle that are the same: 362996

### DIFF
--- a/course_info/static/course_info/js/controllers/SearchController.js
+++ b/course_info/static/course_info/js/controllers/SearchController.js
@@ -261,7 +261,7 @@
                             render: function(data, type, row, meta) {
                                 var url = '#/details/' + row.cid;
                                 var sub_title = '';
-                                if (row.sub_title && row.sub_title != row.description) {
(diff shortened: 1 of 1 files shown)


course_info/static/course_info/js/controllers/SearchController.js
```diff
@@ -260,7 +260,10 @@
                             data: null,
                             render: function(data, type, row, meta) {
                                 var url = '#/details/' + row.cid;
-                                var sub_title = row.sub_title ? ': ' + row.sub_title : '';
+                                var sub_title = '';
+                                if (row.sub_title && row.sub_title != row.description) {
```
sapnamysore: Can we make this a case insensitive check?

sapnamysore:
Looks good 👍

PR #211, opened 2017-06-26
TLT-2898 Truncate title and/or subtitle if beyond length

Chris-Thornton-Harvard:
To prevent the 'Course Details' column in the 'Search Courses' app from becoming too wide, we will truncate the title and/or subtitle if it goes beyond a specified length and append '...' to each.

The maximum amount of characters for a given column is 80.
If there is a title and subtitle, the title can use up to the first 40 of the total before being truncated and the remaining characters will be the subtitle.
If there is a title and no subtitle, the title can use up to 80 characters.

[TLT-2898](https://jira.huit.harvard.edu/browse/TLT-2898)


sapnamysore:
Looks good 👍 ,   validated it on canvas.dev

PR #213, opened 2017-06-26
Course conclusion date picker in 'Search Courses' tool

Chris-Thornton-Harvard:
Using the 'Search Courses' tool, the 'Course conclusion override field' for a course is now an editable field using a Bootstrap date picker. 

The term that the course instance belongs to and the terms conclusion date (if one is set) is displayed next to the conclusion date field.

Demoed to the PO and was given approval.
Currently deployed on DEV for testing.

[TLT-1489](https://jira.huit.harvard.edu/browse/TLT-1489)


course_info/static/course_info/js/directives/HuDatePicker.js
```diff
@@ -0,0 +1,38 @@
...
+                isLoading: '&',
+                label: '@',
+                modelValue: '=',
+                term: '=',
+                termConcludeDate: '=',
+                isSelectedDateInPast: '='
+            },
+            templateUrl: 'partials/hu-date-picker.html',
+            link: function (scope, element) {
+                $('#dp-alert').hide();
```
elliottyates: Could `#dp-alert` become `.dp-alert`, and these linked hide/show actions could point to `$(element).find('.dp-alert')`, similar to what's done for `inputElement`, in order to make this 'reusable' in the sense of not having this directive-scoped code potentially refer to elements outside of its DOM subtree? Would that work?

Chris-Thornton-Harvard:
A call to the controller from the directive will now toggle the date picker alert if an invalid/valid date has been entered.

course_info/templates/course_info/partials/hu-date-picker.html
```diff
@@ -0,0 +1,32 @@
...
+            <span id="span-course-{{field}}" ng-hide="editable">
+                {{modelValue}}
+            </span>
+        </div>
+    </div>
+    <div class="col-sm-8">
+        <div ng-show="{{termConcludeDate}}">
+            This course belongs to <b>{{term}}</b> which is set to conclude on <b>{{termConcludeDate}}</b>.
+        </div>
+        <div ng-hide="{{termConcludeDate}}">
```
sapnamysore: Nitpick: should use ng-if in place of the show/hide, to avoid both elements being rendered

sapnamysore:
Looks good 👍 .  Noticed one thing while testing that's a UI suggestion. (this is regarding submitting after a date alerts is shown and the user possibly not noticing that the old date is saved - will check with PO and open an enhancement if required). 
Good with this PR other than one minor thing.

PR #216, opened 2017-07-21
TLT-3150 Change conclude date time

Chris-Thornton-Harvard:
[TLT-3150](https://jira.huit.harvard.edu/browse/TLT-3150)

Changed the conclude time from 11:59PM to 12:01AM.


eparker71:
👍 Looks great!!!

PR #218, opened 2017-07-26
Select all checkbox functions properly across pages

Chris-Thornton-Harvard:
Fixes issue where the select all checkbox was selecting the first X records from the unsorted data rather than selecting the first X records from the sorted data.
The top and bottom select all checkboxes now act uniformly across datatable pages.


eparker71:
👍

PR #219, opened 2017-07-26
Check for checkbox length to be greater than 0

Chris-Thornton-Harvard:
Make sure the "Select All" checkboxes are unselected when the table is empty.

screen shot 2017-07-26 at 3 17 48 pm


eparker71:
👍

PR #221, opened 2017-09-14
TLT-3185 Blueprint Courses

Chris-Thornton-Harvard:
[TLT-3185](https://jira.huit.harvard.edu/browse/TLT-3185)
This adds the newly released feature of creating a Canvas blueprint course.
Accounts selected to be a blueprint course in the 'Canvas Site Creator' will have the BLU- prefix and be created at the school level rather than the ILE or SB sub-accounts.

This is currently waiting on the 0.10.2 release of the Canvas Python SDK located at
https://github.com/penzance/canvas_python_sdk/releases
Until that is released, the update functionality will not work on dev and you will receive an error message.

### DIFF
--- a/canvas_site_creator/api.py
+++ b/canvas_site_creator/api.py
@@ -252,9 +252,8 @@ def create_canvas_course_and_section(request):
             )
             try:
                 update_course(**update_parameters).json()
-            except Exception as e:
+            except:
                 logger.exception("Error creating blueprint course via update with request {}".format(update_parameters))
-                logger.exception("Exception details: {}".format(e))
                 return JsonResponse({}, status=500)
     except Exception as e:
         message = u'Error creating new course via SDK with request={}'.format(
--- a/canvas_site_creator/api.py
+++ b/canvas_site_creator/api.py
@@ -10,7 +10,7 @@ from django.views.decorators.http import require_http_methods
 from canvas_course_site_wizard.models import CanvasCourseGenerationJob
 from canvas_sdk.exceptions import CanvasAPIError
 from canvas_sdk.methods.content_migrations import create_content_migration_courses
-from canvas_sdk.methods.courses import create_new_course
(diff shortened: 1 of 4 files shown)


canvas_site_creator/api.py
```diff
@@ -236,8 +238,19 @@ def create_canvas_course_and_section(request):
...
+        if is_blueprint:
+            update_parameters = dict(
+                request_ctx=SDK_CONTEXT,
+                id=course_result['id'],
+                course_blueprint=True
+            )
+            update_result = update_course(**update_parameters).json()
     except Exception as e:
         message = u'Error creating new course via SDK with request={}'.format(
             request_parameters)
```
sapnamysore: The exception block is for both the canvas calls.However the logging is relevant to the create call failing and prints the request_parameters even if there would be an exception with the second call. Maybe you could move the  update_course call into a separate block and log the update_parameters in the error message?

canvas_site_creator/static/canvas_site_creator/js/controllers/CreateNewCourseController.js
```diff
@@ -224,30 +237,37 @@
...
             var data = {
-                departments: [departmentId],
                 registrar_code: $scope.course.code,
                 registrar_code_display: $scope.course.code,
-                school: $scope.school.id,
+                school: $scope.school.id
             };
+            // Only include departments if this is not a blueprint course
+            if(!$scope.isBlueprint) {
+                data.departments = [departmentId]
```
sapnamysore: missing semicolon at the end of the line?

Chris-Thornton-Harvard: Good catch!

canvas_site_creator/api.py
Chris-Thornton-Harvard: I added a nested try except to log the update parameters and exception details if there is an error during the course update.

PR #222, opened 2017-09-25
Changed ng-show to ng-if for blueprint option

Chris-Thornton-Harvard:
Safari was not properly handling ng-show for the BLU option in the dropdown menu. This caused BLU to be a selectable option outside of checking the box. This changed the ng-show for that option to ng-if.
Currently deployed on dev.
Tested on Chrome, Firefox and Safari.


sapnamysore:
👍 . verified the deployed fix on safari and FF

PR #223, opened 2017-11-27
TLT-3189 Conclude and unconclude individual enrollments in a course

Chris-Thornton-Harvard:
[TLT-3189](https://jira.huit.harvard.edu/browse/TLT-3189)

Adds the ability to conclude and unconclude individual enrollments in a course by using a date picker in the 'Conclusion Date Override' column.

Currently deployed on dev
https://canvas.dev.tlt.harvard.edu/accounts/15/external_tools/359
I search using all default filters in GSE and select the first course, which should have a few current enrollments.

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -4,7 +4,7 @@
 
     function PeopleController($scope, angularDRF, $compile, courseInstances,
                               djangoUrl, $http, $log, $q, $routeParams,
-                              $uibModal, $window) {
+                              $uibModal, $window, $filter) {
         // set up constants
         $scope.sortKeyByColumnId = {
             0: 'name',
@@ -65,6 +65,7 @@
             $scope.tracking.failures++;
             return null;
         };
+
         $scope.addNewMemberToCourse = function(userPostParams, userName,
                                                searchTerm) {
             /* Adds a single person to the course; handles the actual posting of
@@ -118,6 +119,7 @@
                 .then(handlePostSuccess, handlePostError)
                 .finally($scope.updateProgressBar);
         };
+
         $scope.addPeopleToCourse = function(searchTermList) {
 
             /* looks up HUIDs, XIDs, and/or email addresses from searchTerms
@@ -174,9 +176,11 @@
             };
             $scope.removeFailure = null;
         };
+
         $scope.closeAlert = function(source) {
             $scope.messages[source] = null;
         };
+
         $scope.compareRoles = function(a, b) {
             /*
              * we want to sort roles by a combination of two fields.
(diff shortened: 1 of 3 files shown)


course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -531,26 +560,218 @@
...
+                })
+        };
+
+        // Updates the current course instances 'last_updated' field to be the current date and time.
+        $scope.updateCourseInstanceLastUpdated = function() {
+            var url = djangoUrl.reverse('icommons_rest_api_proxy',
+                                        ['api/course/v2/course_instances/'
+                                         + $scope.courseInstanceId + '/']);
+            $http.patch(url);
+        };
```
sapnamysore: This line had me verify if the update_date on course_enrollee/staff/student gets updated when the conclude_date is modified by the  rest api call on line 748, and it doesn't seem to be. .  I made a comment on the other PR that is closed, just cross referencing it here
https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/114/files#diff-2de6c0d199fd14c7c2ed343274ad367fR1321

sapnamysore:
Other than the suggestion to change the var names and possibly fixing the update_date, looks good!

Some notes: 
- Tested it on Safari and Chrome and looks good in both browsers
- Unsetting a date looks a bit trickier (compared to setting it)  but works when you click outside.  I was trying to clear out the date field and press enter, but that retains the original date. Clicking outside the input area after clearing the date works.

course_info/static/course_info/js/controllers/PeopleController.js
Chris-Thornton-Harvard: This call will only update the last_updated field on a course instance. 
I will add a trigger for update on the enrollee/staff/student so the update_date field gets updated.
I am going to redeploy my branch to dev as the commit containing this call was added after my last deploy.

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -531,26 +560,218 @@
...
+                    if ($scope.isSelectedDateInPast(selectedDate)) {
+                        // Set a .1 second delay before displaying error message.
+                        // When the datepicker is closed, it creates a click event which would close this display
+                        // before it is rendered.
+                        setTimeout(
+                            function() {
+                              $scope.addPopOverToCell(userID, 'failure', 'You can only pick a date in the future.');
+                            }, 100)
+                    } else {
+                        var roleID = $scope.peopleData[userID]['role']['role_id'];
```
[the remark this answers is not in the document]
Chris-Thornton-Harvard: The peopleData dictionary is storing information directly from the REST API call requiring the use of 'role_id'.

Chris-Thornton-Harvard:
Thanks for reviewing this!

PR #225, opened 2018-01-18
Cleaned up javascript hack that was added in the past to render Teach…

sapnamysore:
…ing Fellow as TA per details in JIRA issue https://jira.huit.harvard.edu/browse/TLT-3273

Attaching some screenshots for reference:
a) This is with the JS hack(on dev)
screen shot 2018-01-17 at 6 02 43 pm

b) This is from my local instance
screen shot 2018-01-17 at 6 01 57 pm
, with the fix

C)DB setup
screen shot 2018-01-17 at 6 02 59 pm

**Note**:  once this fix is rolled out, we can adjust the values/labels for TF and TA roles in the user_role table   

@Chris-Thornton-Harvard , @cmurtaugh

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -247,11 +247,6 @@
             // creates a new remove user confirmation modal, and
             // stashes this membership object on the modal's child scope.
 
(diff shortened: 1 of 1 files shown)


cmurtaugh:
:+1: looks good to me!

PR #227, opened 2018-02-22
Hotfix/v1.20.1

bermudezjd:
I'm going to test the `psycopg2` update on dev...


sapnamysore:
👍

cmurtaugh:
👍  looks good to me -- just wonder if we should add the `--no-binary` flag for psycopg2

bermudezjd:
I think I would do so for subsequent versions, but I did some digging and the binary for 2.7.4 was built against a slightly newer version of OpenSSL than what Xenial comes with (1.0.2n vs. 1.0.2g) so I'd rather leave it for now.   If the version had been older, I'd definitely install from source and we should do so starting at 2.8 anyways.

PR #233, opened 2018-05-04
TLT-3359 Fix Canvas site creator

Chris-Thornton-Harvard:
The "Associated Site" column was not rendering the url to the newly created site


Chris-Thornton-Harvard:
Thanks!

PR #236, opened 2018-05-10
Upgrade CAAT Django to 1.11

Chris-Thornton-Harvard:
Required an update of the django-angular versioning.
Currently deployed on dev.
Jamine tests all pass.


sapnamysore:
👍Looks good!
We should test all the tools thoroughly

PR #239, opened 2018-06-08
Task/sapnmysore/site creator load issue

sapnamysore:
Changes to improve the performance of Site creator screens that were timing out.
Deployed on dev and been tested for both ILE and Bulk sites


cmsnowden:
Looks good.to me.

PR #244, opened 2018-08-06
Audit Details page

Chris-Thornton-Harvard:
Moved the revert button to the audit details page .
Pulled in Carters work on the audit page from his branch and updated accordingly


cmsnowden:
Looks good as far as pulling in what was there. We need to fill put the details and some other things (e.g. if we try to add a new job after having just added one, it fails because we've lost the sis_account_is associated with the school.

Chris-Thornton-Harvard:
Thanks! I added a few sub-tasks to track those issues

PR #247, opened 2018-08-09
a bunch of UI tweaks, and some renaming of things

cmurtaugh:
* tweaked breadcrumb text throughout
* linked breadcrumb on the detail screen so user can get back to the list 
* added `created` column to list view
* added a tooltip on the refresh-screen button
* on the detail screen, added some text to explain why the revert button is disabled
* on the detail screen, formatted the top tables using the bootstrap `table` class
* added a Cancel button to the create new job form
* made the buttons line up with the input fields on the create new job form
* a few other minor UI tweaks

I changed the names of some view classes and templates: 
* bulk_settings_job_audit.html --> job_detail.html
* bulk_settings_list.html --> job_list.html
* BulkSettingsAuditView --> BulkSettingsDetailView
* create_new_setting --> create_new_job

In `settings/base.py` I changed the timezone from `UTC` to `America/New_York` -- this is necessary to get timestamps to display correctly in the local timezone when they're rendered in views.  Note that timestamps are still stored in UTC in the database -- this is just for rendering.

I changed the length of the `setting_to_be_modified` field to 50 since we were already exceeding the prior limit of 20.


Chris-Thornton-Harvard:
👍 Looks good!

PR #256, opened 2018-12-12
For Bulk Site Creator, exclude any course instances that have a paren…

sapnamysore:
…t_course_instance_id populated.

Deployed on dev: https://canvas.dev.tlt.harvard.edu/accounts/5/external_tools/359
 Tested with **Divinity school, Spring 2016-17**, it was rendering 118 courses before the change, now has 117 after I explicitly set one of the courses parent value(course_instance_id=379439)


Chris-Thornton-Harvard:
👍 Looks good!

PR #259, opened 2019-02-08
TLT-3591 Angular reverse issue fix

Chris-Thornton-Harvard:
This installation of CAAT, currently deployed on dev, uses the XML from this branch's tool config
https://canvas.dev.tlt.harvard.edu/accounts/10/external_tools/521

Fixes issue where there was an error in the site creator during a Angular reverse call


sapnamysore:
👍 . Tool looks good on dev.

PR #260, opened 2019-02-11
TLT-3595 Added parent_course_instance to patch data in CreateNewCourseController

Chris-Thornton-Harvard:
The serializer that validates patch calls for a course instance requires parent_course_instance to be included in the call. The CreateNewCourseController is the only spot requiring this update as we already pass this value in the course info controller.


Chris-Thornton-Harvard:
Closed and rebranched off of develop in same named branch as this one

PR #282, opened 2019-10-30
TLT-3742 ILE/SB Term Dropdown fix

Chris-Thornton-Harvard:
Add a limit of 100 when getting the list of terms for the dropdown and sort by the terms end date


cmurtaugh:
:+1: Looks good to me -- I'll check it out on canvas.dev

PR #286, opened 2019-11-18
Fixes the Angular Reverse Issue

Chris-Thornton-Harvard:
Made some changes to the URL patterns from path to re_path to handle the query string parameters during the Angular reverse process


cmurtaugh:
:+1:

PR #291, opened 2020-01-16
Task/thornton/tlt 3772/xlisting exceptions

Chris-Thornton-Harvard:
- The validate_inputs method now does not raise an exception but rather returns a Bool determining if the inputs are valid or not.
- Optimized the imports
- Added some logging of the exception trace


sapnamysore:
Looks good to me 👍

PR #292, opened 2020-01-22
TLT-3689 Check if department exists prior to creating course

Chris-Thornton-Harvard:
The logic for this change is in DCCSW.


sapnamysore:
The other 2 PRS are approved. You could create releases of SDK and dccsw and repoint to the  right version of libraries..

I am trying to validate it on qa.

sapnamysore:
Tested on qa. Looks good 👍

PR #294, opened 2020-02-19
Task/thornton/tlt 3793/add logging site deletion

Chris-Thornton-Harvard:


sapnamysore:
Looks good. 👍 .
Had some trouble validating it due to the env changes we are doing. But on deleting a properly populated site, was able to validate on QA

PR #296, opened 2020-02-25
Task/sapnamysore/tlt 3713/fix conclude date

sapnamysore:
1. Modified  CAAT's course search screen and the people tab (both of which have conclude date widgets)  to save tomorrow's date and the validation is fixed.
2. Also fixed the formatting in order to save  12:01Am of current date. But this is in UTC(which is correct) , we would also need to modify the DB columns to reflect timezones. Creating a new story to fix the database columns (more info in the Jira ticket) 
[Deployed on qa, testable on canvas.devops]

Depends on: https://github.com/Harvard-University-iCommons/icommons_rest_api/pull/137

### DIFF
--- a/course_info/static/course_info/js/controllers/DetailsController.js
+++ b/course_info/static/course_info/js/controllers/DetailsController.js
@@ -334,7 +334,6 @@
                     patchData[field] = dc.formDisplayData[field];
                 }
             });
-            console.log(url);
 
             $http.patch(url, patchData)
                 .then(function finalizeCourseDetailsPatch(response) {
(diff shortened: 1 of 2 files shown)


course_info/static/course_info/js/controllers/DetailsController.js
```diff
@@ -335,6 +334,8 @@
                     patchData[field] = dc.formDisplayData[field];
                 }
             });
+            console.log(url);
```
Chris-Thornton-Harvard: Log statment

Chris-Thornton-Harvard:
👍 Looks good!

PR #302, opened 2020-08-27
Stripped any spaces from  primary and secondary input

sapnamysore:
Handled any spaces being input for the ids. (https://jira.huit.harvard.edu/browse/TLT-3880)

### DIFF
--- a/cross_list_courses/utils.py
+++ b/cross_list_courses/utils.py
@@ -74,6 +74,8 @@ def create_crosslisting_pair(primary_id, secondary_id, request):
     try:
         # Validate the primary and secondary values before processing.
         # Validation messages are transmitted to the view by this method
+        primary_id = primary_id.strip()
+        secondary_id = secondary_id.strip()
         is_valid = validate_inputs(primary_id, secondary_id, request)
         
         if is_valid:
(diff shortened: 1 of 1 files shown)


cross_list_courses/utils.py
```diff
@@ -273,8 +275,8 @@ def _update_canvas_cross_listing(primary_sis_id, secondary_sis_id, request):
     # course. Neither of these sections should exist in the former
     # primary course (though there may still appear to be manually
     # created sections there that we will have to clean up manually).
-    secondary_section_id = 'sis_section_id:{}'.format(secondary_sis_id)
-    primary_course_id = 'sis_course_id:{}'.format(primary_sis_id)
```
cmurtaugh, line 277: Hey @sapnamysore -- sorry I didn't catch this sooner, but the `.strip()` in these two lines is going to be applied _after_ the string format happens, rather than to just the `secondary_sis_id` and `primary_sis_id` values. This means that if those values have trailing spaces they will be trimmed, but if they have leading spaces they won't. 

A better option would be like: 
```
secondary_section_id = 'sis_section_id:{}'.format(secondary_sis_id.strip())
```
or even better: 
```
secondary_section_id = 'f'sis_section_id:{secondary_sis_id.strip()}'
```

sapnamysore: No worries @cmurtaugh.
So  the actual stripping of spaces from the input is handled here in  https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/302/files/1d42deabb2c18c9ee4183d04f20fc92ffe869fd2#diff-7a393d24483281978eb1c0079ff5feadR77 that cleeans up the input before invoking this update method on line 114 (https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/302/files/1d42deabb2c18c9ee4183d04f20fc92ffe869fd2#diff-7a393d24483281978eb1c0079ff5feadR114)

This is an additional check that's probably not needed. I had tested it on dev and re-tried just now with spaces before and after and it handles it correctly..https://canvas.dev.tlt.harvard.edu/accounts/10/external_tools/2

Would you rather I remove the check on 276? or update it with the suggested logic?

cmurtaugh:
Ah, gotcha! I think I’d just remove it altogether if it’s being handled upstream. Less code to worry about going forward! :-) Obviously not a high priority to remove it since it works ok now.
 
—Colin

PR #306, opened 2021-01-29
TLT-3967  Update DCCSW version to configure install_requires

Chris-Thornton-Harvard:


Chris-Thornton-Harvard:
Once DCCSW is merged and tagged, I will update the requirements file

PR #307, opened 2021-02-09
Feature/masquerade tool

sapnamysore:
Initial version of  masquerade tool.

### DIFF
--- a/masquerade_tool/templates/masquerade_tool/index.html
+++ b/masquerade_tool/templates/masquerade_tool/index.html
@@ -1,48 +1,44 @@
 {% extends 'masquerade_tool/base.html' %}
-
-{% block body %}
 {% load static %}
 
-<nav>
-  <h3>
-    <a href="{% url 'dashboard_account' %}">Admin Console</a>
-    <small><i class="fa fa-chevron-right"></i></small>
-    Masquerade Tool
-  </h3>
-</nav>
+{% block body %}
+  <nav>
+    <h3>
+      <a href="{% url 'dashboard_account' %}">Admin Console</a>
+      <small><i class="fa fa-chevron-right"></i></small>
+      Masquerade Tool
+    </h3>
+  </nav>
 
-<body role="application">
-    <div class="container-fluid">
+  <div class="container-fluid">
     <br><br>
     <div class="row">
-        <div class="col-sm-6">
-            <p>Click the button below to temporarily assume the Masquerade role. Once granted, you'll be able to use Canvas
-                "<a href="https://community.canvaslms.com/t5/Admin-Guide/How-do-I-act-as-another-user-in-an-account/ta-p/161"
-                target="_blank">Act As User</a>" function for {{ session_mins }} minutes.
-            </p>
-        </div>
+      <div class="col-sm-6">
+        <p class="lead">Click the button below to temporarily assume the Masquerade role. Once granted, you'll be able to use
+          Canvas'
+          "<a href="https://community.canvaslms.com/t5/Admin-Guide/How-do-I-act-as-another-user-in-an-account/ta-p/161"
+            target="_blank">Act As User</a>" function for {{ session_mins }} minutes.
+        </p>
+      </div>
     </div>
     <br>
     <div class="row">
-        <form class="form-inline" method="POST" action="{% url 'masquerade_tool:add_role' %}">{% csrf_token %}
-
-            <div class="col-sm-3">
-                <button type="submit" name="masquerade_req" id="masquerade" class="btn btn-primary" data-loading-text="<i class='fa fa-spinner fa-spin'></i> Requesting...">
-                <i class="fa"></i> Request Masquerade Access</button>
-            </div>
-        </form>
+      <form class="form-inline" method="POST" action="{% url 'masquerade_tool:add_role' %}">{% csrf_token %}
+        <div class="col-sm-3">
+          <button type="submit" name="masquerade_req" id="masquerade" class="btn btn-primary"
+            data-loading-text="<i class='fa fa-spinner fa-spin'></i> Requesting...">
+            <i class="fa"></i> Request Masquerade Access</button>
+        </div>
+      </form>
     </div>
-
-</div>
-</body>
+  </div>
+{% endblock %}
 
 {% block javascript %}
-    {{ block.super }}
-    <script>
-        $('#masquerade').on('click', function () {
-            var $btn = $(this).button('loading')
-        })
-    </script>
+  {{ block.super }}
+  <script>
+    $('#masquerade').on('click', function () {
+      var $btn = $(this).button('loading')
+    })
+  </script>
 {% endblock javascript %}
-
-{% endblock %}
--- a/masquerade_tool/templates/masquerade_tool/index.html
+++ b/masquerade_tool/templates/masquerade_tool/index.html
@@ -1,0 +1,49 @@
+{% extends 'canvas_site_deletion/base.html' %}
+
+{% block body %}
+
+<nav>
+  <h3>
+    <a href="{% url 'dashboard_account' %}">Admin Console</a>
+    <small><i class="fa fa-chevron-right"></i></small>
+    Masquerade Tool
+  </h3>
+</nav>
+
+<body role="application">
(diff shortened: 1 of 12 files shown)


masquerade_tool/templates/masquerade_tool/index.html
```diff
@@ -0,0 +1,43 @@
...
+
+{% block content %}
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-med-9">
+        <br><br>
+        <p class="lead"><strong>Click the button below to temporarily assume the Masquerade role.</strong></p>
+        <p class="lead">Once granted, you'll be able to use
+          Canvas'
+          "<a href="https://community.canvaslms.com/t5/Admin-Guide/How-do-I-act-as-another-user-in-an-account/ta-p/161"
```
elliottyates, line 22: We may want to add `rel="noreferrer"` to this link, per best practices on cross-site links: https://web.dev/external-anchors-use-rel-noopener/

masquerade_tool/templates/masquerade_tool/index.html
```diff
@@ -0,0 +1,43 @@
...
+
+      </div>
+    </div>
+  </div>
+{% endblock %}
+
+{% block javascript %}
+  {{ block.super }}
+  <script>
+    $('#masquerade').on('click', function () {
```
elliottyates, line 39: Is this actually doing anything?

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+@lti_permission_required(settings.PERMISSION_MASQUERADE_TOOL)
+@require_http_methods(['GET'])
+def index(request):
+    context = {'session_mins': MASQUERADE_SESSION_MINS}
+    return render(request, 'masquerade_tool/index.html', context)
+
+@login_required
+@lti_role_required(const.ADMINISTRATOR)
+@lti_permission_required(settings.PERMISSION_MASQUERADE_TOOL)
+@require_http_methods(['GET', 'POST'])
```
elliottyates, line 33: Do we want to remove `'GET'` here?

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+def add_role(request):
+    context = {}
+    try:
+        canvas_user_id = request.LTI['custom_canvas_user_id']
+        logger.info('User {} requested masquerade access at {}'.format(
+            request.LTI['lis_person_name_full'], datetime.now()))
```
elliottyates, line 39: More a question in general about our practices, but do we try to log timezone-aware dates in our log messages, vs. using e.g `utcnow()`?

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+        payload = {
+            'user_id': canvas_user_id,
+            'action': 'add',
+        }
+        logger.debug('payload :{}'.format(payload))
+        result = lambda_client.invoke(
+            FunctionName =FUNCTION_ARN,
+            Payload=json.dumps(payload),
+        )
+        logger.info('lambda result = {}'.format(result))
```
elliottyates, line 51: Maybe consider adding `FUNCTION_ARN` for debugging purposes.

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+        invoke_status = result['ResponseMetadata']['HTTPStatusCode']
```
elliottyates, line 52: Unused? It's a little confusing too, since there's also a var `status` in this function.

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+        exp_dt = None
+        # Check if 'expires' is being sent. If role already exists, expires won't be set.
+        if "expires" in response_payload:
+            exp_dt = dateutil.parser.isoparse(response_payload["expires"])
+
+        logger.info(' payload response =  {}'.format( response_payload))
+        context = {'expiry_time': exp_dt, 'status': status, 'session_mins': MASQUERADE_SESSION_MINS}
+    except Exception as e:
+        logger.error("Error while processing request from {}".format(request.LTI['lis_person_name_full']))
+        logger.error(e)
```
elliottyates, line 65: Worth a `logger.exception` here instead, and use the error message from line 64?

masquerade_tool/views.py
```diff
@@ -0,0 +1,68 @@
...
+        status = response_payload["status"]
+        exp_dt = None
+        # Check if 'expires' is being sent. If role already exists, expires won't be set.
+        if "expires" in response_payload:
+            exp_dt = dateutil.parser.isoparse(response_payload["expires"])
+
+        logger.info(' payload response =  {}'.format( response_payload))
+        context = {'expiry_time': exp_dt, 'status': status, 'session_mins': MASQUERADE_SESSION_MINS}
+    except Exception as e:
+        logger.error("Error while processing request from {}".format(request.LTI['lis_person_name_full']))
```
elliottyates, line 64: I'm not sure whether `request.LTI['lis_person_name_full']` would be guaranteed to exist here, but to be safe we might want to use `request.get('lis_person_name_full', '(list_person_name_full missing from LTI context)')` or something like that to avoid this error being swallowed in that case. Recent pythons can keep track of the error stack, but not sure what magic our app stack and/or Django has for catching and displaying errors, so could be worth avoiding these kinds of possibilities where easy.

masquerade_tool/templates/masquerade_tool/success.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+  <div>
+    <br><br>
+    {% if status == 'role_exists' %}
+    <p class="lead"><strong>You already have the Masquerade role.</strong></p>
+    <p class="lead">
+      You can use the
+      "<a href="https://community.canvaslms.com/t5/Admin-Guide/How-do-I-act-as-another-user-in-an-account/ta-p/161"
+        target="_blank">Act As User</a>" function. The role cannot be granted again until the current session expires.
```
elliottyates, line 21: See other `noreferrer` comment.

masquerade_tool/templates/masquerade_tool/success.html
```diff
@@ -0,0 +1,36 @@
...
+    </p>
+    {% else %}
+    <p class="lead"><strong><i class="fa fa-check-circle text-success"></i> You've been granted the Masquerade role.</strong></p>
+    <p class="lead">
+      It will automatically expire in {{ session_mins }} minutes.
+      You can use the "<a
+        href="https://community.canvaslms.com/t5/Admin-Guide/How-do-I-act-as-another-user-in-an-account/ta-p/161"
+        target="_blank">Act As User</a>" function until {{ expiry_time }}
+    </p>
```
elliottyates, line 30: See other `noreferrer` comment.

masquerade_tool/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path, re_path
```
elliottyates, line 1: `re_path` not used?

elliottyates (approved):
👍

masquerade_tool/views.py
elliottyates: For that matter, if the log has a timestamp anyway, is this a little redundant?

masquerade_tool/templates/masquerade_tool/index.html
sapnamysore: Yeah, it displays the spinner on submission, and disables the button

elliottyates: Ah, this must be some bootstrap/jquery ✨ magic ✨ . I can't find any documentation for it at first glance (this `button('loading')` feature) but 👍 if it works! Thanks.

masquerade_tool/views.py
sapnamysore: Yeah, it's redundant. I think I added it initially as it maybe  useful for  client services as they sometimes use splunk

masquerade_tool/views.py
sapnamysore: Good point

sapnamysore:
Thanks. Made some changes and validated on QA. Merging as it is already approved

PR #312, opened 2021-04-19
Added search_courses permission

cmurtaugh:
This PR adds a custom permission setting `search_courses` to control visibility of the Search Courses dashboard card, and to control access to the views in the `course_info` tool. 

This enables us to grant access to the dashboard itself without also granting access to any specific tools. 

Resolves TLT-4000.

### DIFF
--- a/canvas_account_admin_tools/settings/base.py
+++ b/canvas_account_admin_tools/settings/base.py
@@ -396,15 +396,14 @@ ICOMMONS_REST_API_TOKEN = SECURE_SETTINGS.get('icommons_rest_api_token')
 ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = False
 
 PERMISSION_ACCOUNT_ADMIN_TOOLS = 'account_admin_tools'
+PERMISSION_SEARCH_COURSES = 'search_courses'  # aka course_info
 PERMISSION_PEOPLE_TOOL = 'people_tool'
 PERMISSION_XLIST_TOOL = 'cross_listing'
 PERMISSION_SITE_CREATOR = 'manage_courses'
 PERMISSION_PUBLISH_COURSES = 'publish_courses'
 PERMISSION_BULK_COURSE_SETTING = 'bulk_course_settings'
 PERMISSION_CANVAS_SITE_DELETION = 'canvas_site_deletion'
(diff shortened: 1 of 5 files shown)


canvas_account_admin_tools/settings/base.py
```diff
@@ -396,15 +396,14 @@
 ICOMMONS_REST_API_SKIP_CERT_VERIFICATION = False
 
 PERMISSION_ACCOUNT_ADMIN_TOOLS = 'account_admin_tools'
+PERMISSION_SEARCH_COURSES = 'search_courses'  # aka course_info
```
elliottyates, line 399: Helpful comment, thanks!

canvas_account_admin_tools/views.py
```diff
@@ -1,23 +1,23 @@
...
 from django.views.decorators.http import require_http_methods
-from lti import ToolConfig
-from proxy.views import proxy_view
-
 from django_auth_lti import const
 from django_auth_lti.decorators import lti_role_required
+from lti import ToolConfig
 from lti_permissions.decorators import lti_permission_required
 from lti_permissions.verification import is_allowed
+from proxy.views import proxy_view
```
elliottyates, line 20: Nice, thanks for reorganizing these imports.

canvas_account_admin_tools/views.py
```diff
@@ -88,7 +88,16 @@ def dashboard_account(request):
     custom_canvas_membership_roles = request.LTI['custom_canvas_membership_roles']
 
     """
-    Verify that the curernt user has permission to see the cross listing button
+    Verify that the current user has permission to see the Search Courses (aka course_info) tool
+    """
+    search_courses_allowed = is_allowed(
```
elliottyates, line 93: Is there a place to keep release information, i.e. how we'd like to seed the various databases? Or is that already done in dev, qa, and prod? (Given this is a new permission, I would think it's safe to include it anytime before release, right?)

elliottyates (approved):
👍

canvas_account_admin_tools/views.py
cmurtaugh: We don't have any established place to keep that kind of information. I'll take care of updating the database before deployment.

PR #317, opened 2021-07-15
TLT-4033: Allow Canvas Site Creator to ignore courses that are set to be synced to canvas

PatrickMTonne:
This PR works to resolve a bug where creating a course via Canvas Site Creator (within CAAT) could occasionally return an unexpected error due to potential clashing with the background async Canvas feed process. A more detailed description of this bug can be found in the associated [JIRA ticket](https://jira.huit.harvard.edu/browse/TLT-4033).

To resolve this issue, the following updates were made:

- Set the `create_job` view to exclude courses from its query where `sync_to_canvas=1` in the case that "Create All" is `True` (i.e. the user has not selected specific courses from the selection menu; they want sites created for all listed courses)(9342bbc). The logic associated with the other scenario (i.e. the user selects just a subset of listed courses) did not need updating.
- Ensure that the courses listed in the course selection menu are only those that are not set to be synced to Canvas (db query update in 87c52ae; template update in 22e4309)
- Within the `index` template, add an additional bullet that lists how many courses without a Canvas site are eligible for site creation via Canvas Site Creator (e.g. "7 courses eligible for site creation via Canvas Site Creator (0 iSites, 0 external)") (template update in 22e4309; Django model update in e45a056; Angular CourseInstanceModel update in a1f05e7).

I tested these changes locally within QA using [this sub-account](https://canvas.qa.tlt.harvard.edu/accounts/11) (nothing special here; it was a fairly arbitrary choice to use this sub-account). Most recent courses already have Canvas sites (as expected), so you'll need to set "Academic Term" to an earlier date to find courses without sites (I generally used "2020 Fall / Full Term").

### DIFF
--- a/canvas_site_creator/api.py
+++ b/canvas_site_creator/api.py
@@ -352,9 +352,8 @@ def course_instances(request, sis_term_id, sis_account_id):
         order_by_operator = '-' if sort_dir == 'desc' else ''
         query_set = query_set.order_by(order_by_operator + COURSE_INSTANCE_DATA_FIELDS[sort_index])
 
-        # fixes TLT-1570 where courses that already have a Canvas course were showing up
-        # in the create list
-        query_set = query_set.exclude(sync_to_canvas=1)
+        # Exclude courses that are automatically fed to Canvas and already have a Canvas site
+        query_set = query_set.exclude(sync_to_canvas=1).exclude(canvas_course_id__isnull=False)
         result = get_course_instance_summary_data(query_set)
 
         data = []
--- a/canvas_site_creator/api.py
+++ b/canvas_site_creator/api.py
@@ -354,8 +354,7 @@ def course_instances(request, sis_term_id, sis_account_id):
 
         # fixes TLT-1570 where courses that already have a Canvas course were showing up
         # in the create list
-        query_set = query_set.exclude(canvas_course_id__isnull=False)
-
+        query_set = query_set.exclude(sync_to_canvas=1)
         result = get_course_instance_summary_data(query_set)
 
         data = []
--- a/canvas_site_creator/templates/canvas_site_creator/index.html
+++ b/canvas_site_creator/templates/canvas_site_creator/index.html
@@ -81,6 +81,23 @@
                             {% verbatim %}
                             {{ courseInstanceModel.totalCoursesWithoutCanvasSiteWithExternal }} external)
                             {% endverbatim %}
+                            {% comment %} <br>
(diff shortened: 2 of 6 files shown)


canvas_site_creator/api.py
```diff
@@ -354,8 +354,7 @@ def course_instances(request, sis_term_id, sis_account_id):
 
         # fixes TLT-1570 where courses that already have a Canvas course were showing up
         # in the create list
-        query_set = query_set.exclude(canvas_course_id__isnull=False)
-
+        query_set = query_set.exclude(sync_to_canvas=1)
```
elliottyates, line 357: I wonder if there would be a case where `sync_to_canvas=0` but `canvas_course_id` is not `null`, e.g. because of some other edge case we can't imagine. Would it be worth ensuring both possibilities are `exclude`d? i.e. ensure it matches the logic in the [`create_job` view](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/317/files#diff-e04e401071cebc10a9425bd6d0e47345d1c6b880fc7dadb37596edcd7e4de409R323-R324)?

elliottyates: We might also consider making the code comment here more generic, and not referencing that older ticket anymore.

canvas_site_creator/templates/canvas_site_creator/index.html
```diff
@@ -81,6 +81,23 @@ <h3>
                             {% verbatim %}
                             {{ courseInstanceModel.totalCoursesWithoutCanvasSiteWithExternal }} external)
                             {% endverbatim %}
+                            {% comment %} <br>
+                            <button type="button" class="btn btn-link pull-right" id="createSitesButton" ng-disabled="createDisabled()" ng-click="handleCreate($event)" data-href="{% url 'canvas_site_creator:course_selection' %}">Continue</button> {% endcomment %}
```
elliottyates, line 85: If this is commented, should we just remove it now?

canvas_site_creator/templates/canvas_site_creator/index.html
```diff
@@ -81,6 +81,23 @@ <h3>
...
+                        </li>
+                        <li>
+                            <ng-pluralize count="courseInstanceModel.totalCoursesWithoutCanvasSiteAndSyncToCanvasFalse"
```
elliottyates, line 88: You mention "add an additional bullet" -- I don't see that in [Admin Console QA](https://canvas.qa.tlt.harvard.edu/accounts/11/external_tools/3), so I suppose this isn't deployed yet outside of a local install?

elliottyates (approved):
👍 As always, thanks for the nice PR description.

PR #328, opened 2022-06-24
Rename the Procfile to Procfile-cronserver so it is picked up in deploy script

Chris-Thornton-Harvard:
Originally the jobs listed in the the Procfile would be created on the Django app-server during a EC2 startup or deploy, as specified in the Django deploy Ansible playbook. With the recent change of the frontend to ECS, that deploy playbook no longer runs and the jobs are not initiated.

This change would modify the name of the Procfile to be Procfile-cronserver so that the jobs are spun up on the Django cron-server
See the following Ansible task for more info:
https://github.com/Harvard-University-iCommons/tlt-ops/blob/master/ansible/roles/django_project_deploy/tasks/procfile-cronserver.yml

Deployed and tested on QA:
https://canvas.qa.tlt.harvard.edu/accounts/5/external_tools/3


cmurtaugh:
:+1:

PR #332, opened 2022-08-07
Cross-listing tool: wider primary / secondary course fields

cmurtaugh:
This PR increases the width of the primary and secondary course fields on the "add new pair" screen, and increases the truncation length of the sub_title field from 40 to 75 characters. 

The code now uses `django.utils.text.Truncator` to perform the truncation which only includes the ellipsis if the string was actually truncated.

Here are a couple of search strings for some HMS courses with longer sub-titles: `su 600m` , `ob 600`.

Closes TLT-4287.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ secure.*
 # files generated via `runserver_plus` CLI command
 *.crt
 *.key
+.python-version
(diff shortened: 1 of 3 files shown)


PatrickMTonne (approved):
👍

PR #334, opened 2022-08-22
Removes the web process from the Procfile

Chris-Thornton-Harvard:
Not needed as process has since moved to ECS
Deployed on QA


cmurtaugh (approved):
👍

Chris-Thornton-Harvard:
Thanks @cmurtaugh !

PR #341, opened 2022-10-18
Task/haydn9000/tlt 4279/list and edit self enroll enabled courses

haydn9000:
- Self enrollment list section (only shows courses for launch school you're in)
- Updated how self enrollment distributable link shows (no resource id at the end)
- Updated role id to show as role name on pages

### DIFF
--- a/canvas_account_admin_tools/settings/base.py
+++ b/canvas_account_admin_tools/settings/base.py
@@ -456,8 +456,6 @@ ADD_PEOPLE_TO_COURSE_ALLOWED_ROLES_LIST = [0, 1, 2, 5, 6, 7, 9, 10, 11, 12, 14, 
 
 SELF_ENROLLMENT_TOOL_ROLES_LIST = SECURE_SETTINGS.get('self_enrollment_roles', [0,10,14])
 
-SELF_ENROLLMENT_TOOL_ROLES_LIST = SECURE_SETTINGS.get('self_enrollment_roles', [0,10,14])
-
 BULK_COURSE_CREATION = {
     'log_long_running_jobs': True,
     'long_running_age_in_minutes': 30,
--- a/canvas_account_admin_tools/settings/base.py
+++ b/canvas_account_admin_tools/settings/base.py
@@ -79,6 +79,7 @@ INSTALLED_APPS = [
     'bulk_course_settings',
     'canvas_site_deletion',
     'masquerade_tool',
+    'self_enrollment_tool',
     'rest_framework',
     'watchman'
 ]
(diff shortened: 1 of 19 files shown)


canvas_account_admin_tools/settings/base.py
```diff
@@ -456,6 +456,8 @@
 
 SELF_ENROLLMENT_TOOL_ROLES_LIST = SECURE_SETTINGS.get('self_enrollment_roles', [0,10,14])
 
+SELF_ENROLLMENT_TOOL_ROLES_LIST = SECURE_SETTINGS.get('self_enrollment_roles', [0,10,14])
```
Chris-Thornton-Harvard, line 459: Dupe of line 457

haydn9000: Thanks! Now updated

self_enrollment_tool/views.py
```diff
@@ -176,3 +256,20 @@ def enroll (request, course_instance_id):
...
+    try:
+        self_enrollment_course = SelfEnrollmentCourse.objects.get(pk=pk)
+        self_enrollment_course.delete()
+    except Exception as e:
+        msg = f'Unable to remove self enrollment course {pk}.'
+        logger.exception(msg)
+        logger.exception(e)
+        messages.error(request, msg)
+
+    return redirect('self_enrollment_tool:index')
```
Chris-Thornton-Harvard, line 275: Notes from our Zoom conversation:
- Rename view to be a little clearer
- Change permission for lti_permission_required to be PERMISSION_SELF_ENROLLMENT_TOOL
- Add exception for object DoesNotExist and return a warning message
- Add logging statement for deleting SelfEnrollmentCourse
- Remove logger.exception(e) as it is redundant to the exception log above it
- Make the HTTP method DELETE

self_enrollment_tool/views.py
```diff
@@ -123,10 +211,8 @@ def enable (request, course_instance_id):
     logger.debug(request)
     role_id = request.POST.get('role_id')
 
-    # todo: retrive role name. Note: Thsi is implemented in teh other story for thsi tool but will remove this comment
-    #  after verifying on merging upstream
-    role_name = request.POST.get('role_id')
-    logger.debug(f'Selected Role_id {role_id} for Self Enrollment in  course {course_instance_id}')
+    role_name = request.POST.get('role_name')
```
cmurtaugh, line 214: Here we probably want to be using the role ID rather than the name.

self_enrollment_tool/views.py
```diff
@@ -176,3 +259,32 @@ def enroll (request, course_instance_id):
         'abort': False,
     }
     return render(request, 'self_enrollment_tool/lookup.html', context)
+
+
+@login_required
+@lti_role_required(const.ADMINISTRATOR)
+@lti_permission_required(settings.PERMISSION_SELF_ENROLLMENT_TOOL)
+@require_http_methods(['DELETE'])
+def delete_self_enroll_course(request, pk):
```
cmurtaugh, line 268: I'd change this to `disable` so that it matches the name of the `enable` view.

cmurtaugh (changes requested):
A few comments.

self_enrollment_tool/views.py
haydn9000: We are also getting role ID in a request above. Let me know if we want to change that

PR #343, opened 2022-11-02
TLT-4228: Enroll users via self-reg link

PatrickMTonne:
Resolves [TLT-4228](https://jira.huit.harvard.edu/browse/TLT-4278).

### High-level update

This PR adds functionality to enroll users via self-reg into the designated Canvas course.

### More granular updates to accomplish the above

Model:
- New UUID field to uniquely identify the self-reg record. This update allows us to remove the currently existing inclusion of `course_instance_id` in the self-reg URL.
    - A few additional migrations files were added in order to make this update. This is due to how Django handles the addition of unique fields to existing models. See this [Slack thread](https://tlt.slack.com/archives/C02F4DY6H/p1666273695400949) for a bit more info.
- Added a `unique_together` constraint on `course_instance_id` and `rold_id` within the `SelfEnrollmentCourse` model. This is to better represent the expected data model, whereby a given course instance my have multiple self-reg URLs for different role types.

Views:
- `/enable`
    - Updated how the self-reg URL is generated. This is to incorporate the `uuid` mentioned above.
- `/lookup`
    - Added role names to the dropdown list.
- `/enroll`
    - This view validates a few things before adding the user to the appropriate enrollment table within the CourseManager DB, then enrolls them in the course via the Canvas API.
    - The UI where we present the self-reg link is pretty bare bones right now and is meant to act merely as an MVP.

### Testing
This branch has been deployed to QA.

I primarily used Sapna's test course (SIS ID 364060). You can use any course for testing, however, but it does need to have a `canvas_course_id` assigned to it within the `course_instance` table.

To test:
- Navigate to the self-reg tool within CAAT.
- Click "Add New"
- Input an SIS ID (you can use the above ID if in QA).
- Select a role type from the dropdown on the following page.
- On "enable," this will either generate a new URL or hand back an already-existing URL (if generated previously).
- Click on the resulting URL to enroll yourself in the course.
    - NOTE: after testing (or maybe as a part of testing), you'll need to remove yourself from the Canvas People page as well as the appropriate enrollment table in `coursemanager`.

### TODOs
- I've noticed that navigating to the self-reg URL in an Incognito tab (not logged into Canvas) leads to a permission error. With the `@login_required` decorator applied to the `/enroll` view, I imagined that it'd send you to the HKey login page, but that doesn't seem to be the case. Any help here is appreciated.
- Some of the log messages are pretty long. It might be worth aligning on some code-formatting standard to improve readability within the IDE (using something like `black` for example).


haydn9000 (approved):
👍

PR #344, opened 2022-11-16
TLT-4336: Separate Search and Enable sections within UI

PatrickMTonne:
Resolves [TLT-4336](https://jira.huit.harvard.edu/browse/TLT-4336).

This PR updates the UI such that after a successful course lookup (via SIS ID), the search box is removed from the subsequent view. Previously, the search box would still be present alongside the resulting course information.

Deployed to QA.


sapnamysore:
Looks good to me! 👍

PR #345, opened 2022-11-18
Task/haydn9000/tlt 4339/confirmation dialog before deleting

haydn9000:
- Added delete confirmation modal
- Sort table by most recently modified

### DIFF
--- a/self_enrollment_tool/templates/self_enrollment_tool/index.html
+++ b/self_enrollment_tool/templates/self_enrollment_tool/index.html
@@ -29,6 +29,26 @@
     <div class="row" style="margin-top: 1em;">
         <a href="{% url 'self_enrollment_tool:add_new' %}" class="btn btn-primary pull-right">Add New</a>
     </div>
+
+    <!-- Modal Delete Confirmation START -->
+    <div class="modal fade" id="DelConfirmModalCenter" tabindex="-1" role="dialog" aria-labelledby="DelConfirmModalCenterTitle" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-centered" role="document">
+            <div class="modal-content">
+                <div class="modal-header text-center">
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                    <h4 class="modal-title" id="DelConfirmModalLongTitle">Are you sure you want to delete this self-enrollment link?</h4>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-primary">Delete</button>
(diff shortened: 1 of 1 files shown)


self_enrollment_tool/views.py
```diff
@@ -355,55 +356,24 @@ def disable(request, uuid):
...
-        # course_enrollee.DeletedEnrollee()
-    except Exception:
-        msg = f'Unable to delete user_id:{user_id} from course_instance_id:{course_instance_id}.'
+        msg = f'Unable to delete self-enroll URL for course uuid: {uuid}.'
         logger.exception(msg)
         messages.error(request, msg)
 
-    logger.info(
-        f'Deleted user_id:{user_id} from self-enroll course_instance_id:{course_instance_id}.')
+    msg = f'Successfully deleted self-enroll URL for course (uuid: {uuid}).'
```
sapnamysore: This message should display a more useful  Course Metadata. Rather than uuid, course name/title that is in the first column will be useful for the user

self_enrollment_tool/views.py
```diff
@@ -56,7 +57,7 @@ def index(request):
     course_instance_ids = {crs_id[0] for crs_id in course_info.values_list('course_instance_id')}
 
     self_enroll_course_list = SelfEnrollmentCourse.objects.filter(
-        course_instance_id__in=course_instance_ids)
+        course_instance_id__in=course_instance_ids).order_by('-last_updated')
```
sapnamysore, line 60: This seems right. But when I test this on qa, the new records are showing up at the bottom of the table rather than the top?

self_enrollment_tool/templates/self_enrollment_tool/index.html
```diff
@@ -41,6 +61,7 @@ <h4 class="modal-title bottom-spacing">Self Enrollment Course List:</h4>
                     <th>Role&nbsp;ID</th>
                     <th>Self&nbsp;Enroll&nbsp;URL</th>
                     <th>Modified&nbsp;By</th>
+                    <th>Last&nbsp;Updated</th>
```
sapnamysore, line 64: Is this on qa? I don't see this column

haydn9000: This was hidden and not used, it's now removed

self_enrollment_tool/views.py
haydn9000: This works, I'll redeploy so we can try again when QA is working

self_enrollment_tool/views.py
haydn9000: Updated this to not include the uuid and just be a simple successful message like we talked about 👍

PR #346, opened 2022-11-19
Self un-enroll LTI (1.3) tool

cmurtaugh:
This PR adds a new LTI 1.3 tool that can be installed into a course in order to let enrollees remove themselves from the course.

The tool appears in the right-hand navigation on the course homepage: "Un-enroll from this course". When a user clicks on the link, we check to make sure that self-enrollment is enabled for this course and that the user is actually self-enrolled. If so, the user is presented with a "Leave this course" button.

When the user clicks the button, the tool removes each enrollment with the source "selfenroll" for this user in this course. It also calls the Canvas API to remove each self-enrolled role from the Canvas course. 

See the README for instructions on the LTI 1.3 installation process. This PR includes three new management commands to facilitate the installation process: `generate_lti_key`, `create_tool_config`, and `list_lti_keys`. 

This PR also modifies the `self_enrollment_tool` so that it will install the new unenrollment LTI tool into the Canvas site when an administrator enables self-enrollment for that course.


haydn9000 (approved):
This looks good 👍 I noticed GitHub is warning about not having new lines at the end of the template files though. Not sure how much that matters

cross_list_courses/templates/add_new.html
```diff
@@ -138,7 +138,7 @@ <h3>
             "bInfo" : false,
             "aoColumnDefs": [ { "orderable": false, "targets": [1,3] } ],
             "oLanguage": {
-              "sSearch": "Filter the list below"
+              "sSearch": "Filter the list"
```
sapnamysore, line 141: Was. this text change intended for the cross listing tool ? Or is it a search and replace by-product while fixing the self reg tool?

cross_list_courses/templates/list.html
```diff
@@ -120,7 +120,7 @@ <h4 class="modal-title bottom-spacing">Cross-Listed Courses in Current Terms:</h
             "bInfo" : false,
             "aoColumnDefs": [ { "orderable": false, "targets": 3 } ],
             "oLanguage": {
-              "sSearch": "Filter the list below"
+              "sSearch": "Filter the list"
```
sapnamysore, line 123: Same as above- was this change intended for cross listing tool?

cross_list_courses/templates/add_new.html
cmurtaugh: Whoops -- an overly aggressive search-and-replace! :-)

cross_list_courses/templates/list.html
cmurtaugh: Ditto - I'll back these out

self_unenrollment_tool/views.py
```diff
@@ -0,0 +1,245 @@
...
+def launch(request: HttpRequest):
+    message_launch = get_message_launch(request)
+    # this next line is necessary even if the variable is not used; if get_launch_data() is not called, launch data will not be saved to the session!
+    launch_data = message_launch.get_launch_data()
+    logger.info('self_unenrollment_tool launch', extra={'launch_data': launch_data})
+    return redirect(reverse('self_unenrollment_tool:index', kwargs={'launch_id': message_launch.get_launch_id()}))
+
+
+@require_lti_launch
+def index(request: HttpRequest, launch_id):
```
PatrickMTonne, line 52: Nice use of type hints. Would it be worth adding types to the other arguments in each view for consistency?

canvas_account_admin_tools/settings/base.py
```diff
@@ -326,7 +330,7 @@
     'loggers': {
         'gunicorn': {
             'handlers': ['gunicorn', 'console'],
-            'level': 'INFO',
+            'level': logging.WARNING,
```
PatrickMTonne, line 333: I assume this is to reduce the noisiness of the `gunicorn` logs?

self_enrollment_tool/templates/self_enrollment_tool/add_new.html
```diff
@@ -6,7 +6,7 @@
   <h3>
     <a href="{% url 'dashboard_account' %}">Admin Console</a>
     <small><i class="fa fa-chevron-right"></i></small>
-      <a href="{% url 'self_enrollment_tool:index' %}">Enable Self Enrollment</a>
+      <a href="{% url 'self_enrollment_tool:index' %}">Self Enrollment</a>
```
PatrickMTonne, line 9: 👍

sapnamysore:
Changes looked good to me👍  . The code walkthru was helpful to understand the logic, workflow and the LTI 1.3 details.. Nice work!

PatrickMTonne (approved):
👍 Looks good! Thanks for walking us through this earlier.

self_unenrollment_tool/views.py
cmurtaugh: :+1:

canvas_account_admin_tools/settings/base.py
cmurtaugh: Yeah -- we send tons of gunicorn log events to Splunk but they're not that useful. I think we really only need to know about warnings and above.

PR #347, opened 2022-11-22
Add copy button and simplify enable UX

PatrickMTonne:
Resolves [TLT-4338](https://jira.huit.harvard.edu/browse/TLT-4338) and [TLT-4340](https://jira.huit.harvard.edu/browse/TLT-4340).

This PR:
- Adds a "Copy URL" button next to each self-reg URL to enable easy distribution.
- Updates the URLs in the table to standalone text instead of links to prevent accidental enrollment.
- Redirects the user back to the `index` page after enabling self-enrollment instead of sending them to a separate `enable_enrollment` page.
- Presents either a `success` or `error` message after redirecting the user back to the `index` page.

NOTE:
- The `copy` function does not work in our self-hosted Canvas environments (canvas.dev and canvas.qa) due to the absence of the `clipboard-write` permission within the Canvas iFrame, a permission necessary for use of the modern [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) that `copy` utilizes. However, this permission is present in the Instructure-hosted Canvas instances (Beta, Test, and regular Canvas), so testing was done in Beta and functionality should work as expected in Prod.

Deployed to Dev (accessed through Beta). Example testing account here: https://harvard.beta.instructure.com/accounts/749/external_tools/21863


self_enrollment_tool/templates/self_enrollment_tool/index.html
```diff
@@ -89,9 +89,11 @@ <h4 class="modal-title bottom-spacing">Self Enrollment Course List:</h4>
                         </td>
                         {% comment %} Self Enroll URL column {% endcomment %}
                         <td>
-                            <a href="{{ crs.self_enroll_url }}">
-                                {{ crs.self_enroll_url }}
-                            </a>
+                            <div>
+                                <span id="url">{{ crs.self_enroll_url }}</span>
+                                <br>
+                                <button id="copy" class="btn btn-primary" onclick="copy(event)">Copy URL</button>
```
cmurtaugh, line 95: I think the button is a bit large; we could use an icon instead to make it fit a little better. 

Since we have FontAwesome in our template you can display an icon like this:
```
<i class="fa fa-clone"></i>
```

cmurtaugh (commented):
One comment re: using an icon instead of the copy button. Otherwise looks good!

self_enrollment_tool/templates/self_enrollment_tool/index.html
PatrickMTonne: Sounds good! Will update.

PR #348, opened 2022-11-29
TLT-4346: Prevent lookup of courses outside current sub-account

PatrickMTonne:
Resolves [TLT-4346](https://jira.huit.harvard.edu/browse/TLT-4346).

This PR:
- Adds an additional search constraint in the `lookup` view to remove the user's ability to search for courses outside of the current sub-account
- Unrelated to the story itself, some CSS was added to the delete icon to make the cursor become a pointer on mouseover. This is to make it more clear that the icon is clickable.

Testing:
These changes have been deployed to dev and can be accessed through beta (example sub-account here: https://harvard.beta.instructure.com/accounts/749/external_tools/21863).


cmurtaugh (approved):
👍 looks good! Nice CSS addition!

PR #352, opened 2022-12-20
TLT-4316: Allow dual TA/Student roles for GSD

PatrickMTonne:
Resolves [TLT-4316](https://jira.huit.harvard.edu/secure/RapidBoard.jspa?rapidView=710&projectKey=TLT&view=detail&selectedIssue=TLT-4316).

This PR adds the ability for GSD courses to enroll a given user as both a Student and TA (two separate enrollments). All other functionality remains the same.

Deployed to dev.

### Testing

To test, search for a GSD course within Search Courses. Select the course and enroll a given user twice, once as a Student and again as a TA. Both enrollments should be accepted.

Attempting to do this outside of a GSD course will result in an error (the same error that has always been present when trying to enroll a user twice).

### DIFF
--- a/course_info/static/course_info/js/controllers/PeopleController.js
+++ b/course_info/static/course_info/js/controllers/PeopleController.js
@@ -35,7 +35,7 @@
             var filteredResults = $scope.filterSearchResults(personRecords);
             if (filteredResults.length == 1) {
                 var memberRecordsInCourse = members[filteredResults[0].univ_id];
-                if (angular.isUndefined(memberRecordsInCourse)) {
+                if (angular.isUndefined(memberRecordsInCourse) || $scope.allowDualEnrollment(memberRecordsInCourse, $scope.selectedRole.roleId)) {
                     var name = $scope.getProfileFullName(filteredResults[0]);
                     var postParams = {
                         user_id: filteredResults[0].univ_id,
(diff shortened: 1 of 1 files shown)


haydn9000:
Update to not let users have more than two roles if that's not desired. 
For example if a user already have the role Guest, they could still get the roles of TA and Student on top of their previous role. 
Thanks for going over this with me in our Zoom chat

course_info/static/course_info/js/controllers/PeopleController.js
```diff
@@ -72,6 +72,27 @@
...
+            // Check if enrollment request is for TA/Student roles in GSD.
+            // For this school only, we allow a given HUID to be enrolled
+            // twice if (and only if) they are enrolled as one of the
+            // two eligible roles (TA/Student).
+
+            var eligibleSchool = 'GSD'
+            var eligibleRoleIds = [0, 5]
+            var memberRole = member[0].role.role_id
+
+            return ($scope.getSchool() === eligibleSchool // ensure course is in GSD
```
sapnamysore, line 89: Would the getSchool always be uppercase GSD? If so it's fine, otherwise making this check case insesnsitive maybe a good idea

haydn9000 (approved):
👍 Looks good! Tested in beta in this GSD course SIS ID: 596625

course_info/static/course_info/js/controllers/PeopleController.js
PatrickMTonne: Added `toLowerCase()` to avoid case issues.

PR #355, opened 2023-01-30
Added async_operations app to this project

haydn9000:
Moved [async_operations](https://github.com/Harvard-University-iCommons/django-icommons-common/tree/main/async_operations) Django app from [django_icommons_common](https://github.com/Harvard-University-iCommons/django-icommons-common) to this project (CAAT).

Note: 
- Tested using Publish Courses. Deployed in Dev for testing.
- Updated [django-canvas-course-site-wizard](https://github.com/Harvard-University-iCommons/django-canvas-course-site-wizard) to no longer require async_operations from django_icommons_common. See changes [here](https://github.com/Harvard-University-iCommons/django-canvas-course-site-wizard/compare/master...task/haydn9000/tlt-4363/remove-old-dependency).

### DIFF
--- a/canvas_account_admin_tools/requirements/base.txt
+++ b/canvas_account_admin_tools/requirements/base.txt
@@ -21,7 +21,8 @@ django-rq==2.5.1
 
 git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v3.0#egg=django-icommons-common==3.0
 
-git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@v3.0#egg=django-canvas-course-site-wizard==3.0
+# git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@v3.0#egg=django-canvas-course-site-wizard==3.0
+git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@task/haydn9000/tlt-4363/remove-old-dependency#egg=django-canvas-course-site-wizard==3.0
 
 git+ssh://git@github.com/Harvard-University-iCommons/django-auth-lti.git@v2.1.0#egg=django-auth-lti==2.1.0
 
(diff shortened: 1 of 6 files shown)


cmurtaugh:
This looks good to me; we will need to be very careful about how we deploy this since the `async_operations_process` table already exists. We'll need to make sure that these new migrations don't break the existing table/data.

haydn9000:
Yeah, maybe we can have a Zoom session to deploy this and work through issues if any

Chris-Thornton-Harvard:
Looks good
I pulled this branch locally and ran showmigrations which showed the following
```
async_operations
 [X] 0001_initial
 [X] 0002_process_created_by_user_id
 [X] 0003_process_status_default_blank
```

I then ran makemigrations which created a new migration that altered 3 fields (didn't dig too deep but might be worthwhile to see why those fields changed) I did not apply this migration.

@haydn9000 I think we should be safe but I would run migrate async_operations in dev or QA to see if anything gets applied.
You could also make a small model change and apply a new migration to make sure things work as expected.

canvas_account_admin_tools/requirements/base.txt
```diff
@@ -21,7 +21,8 @@ django-rq==2.5.1
 
 git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v3.0#egg=django-icommons-common==3.0
 
-git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@v3.0#egg=django-canvas-course-site-wizard==3.0
+# git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@v3.0#egg=django-canvas-course-site-wizard==3.0
+git+ssh://git@github.com/Harvard-University-iCommons/django-canvas-course-site-wizard.git@task/haydn9000/tlt-4363/remove-old-dependency#egg=django-canvas-course-site-wizard==3.0
```
Chris-Thornton-Harvard, line 25: Just a reminder to update this to a tagged version of DCCSW

haydn9000:
@Chris-Thornton-Harvard thanks for the review. I'll look into what changed in the fields and see if anything gets applied 👍

haydn9000:
@Chris-Thornton-Harvard Accidental request above, but we can review again together. I might have some questions about the altered fields you noticed when you ran makemigrations

PR #363, opened 2023-02-08
Update permissions library + fix typo

PatrickMTonne:
Resolves [TLT-4381 ](https://jira.huit.harvard.edu/secure/RapidBoard.jspa?rapidView=710&projectKey=TLT&view=detail&selectedIssue=TLT-4381) and [TLT-4370](https://jira.huit.harvard.edu/browse/TLT-4370).

Depends on [this PR](https://github.com/Harvard-University-iCommons/django-canvas-lti-school-permissions/pull/14) (merged) and [this PR](https://github.com/Harvard-University-iCommons/canvas_python_sdk/pull/3) (merged).

This PR (core story work):
- Removes references to the `lti_permissions` library (TLT-4381)
- Replaces the above references with those from `lti_school_permissions` (TLT-4381)
- Fixes a typo in the cross list tool (TLT-4370)

Note that the new `lti_school_permissions` library creates a new permissions table. This is done as a part of a [migration script](https://github.com/Harvard-University-iCommons/django-canvas-lti-school-permissions/blob/a1aa942ff04f3663f58f492539004100b8079e12/lti_school_permissions/migrations/0003_school_permission_custom_data.py#L24) included in `lti_school_permissions`. From a high level, this involves `lti_school_permissions` examining the old `lti_permissions` table, transforming the data into a format compatible with the new table's schema (since the two schemas are indeed different), and inserts them into the table.

This is obviously very helpful, though it's tough to know if every single permission was moved over (particularly if there were any 'ad-hoc' permissions added to the old table that wouldn't be accounted for in the table-to-table migration script). I looked the data over and nothing stuck out as missing, though an additional pair of eyes could be helpful.

In addition to the above, this PR (non-story work)
- Updates most references to `icommons_common` with those of the newer packages (`django-coursemanager`, `harvard_django_utils`, etc.).
- Replaces all project references to [SessionInactivityExpirationRC](https://github.com/Harvard-University-iCommons/django-icommons-common/blob/a3a2d6e6bd92a007487b657b467449618cb32f75/icommons_common/canvas_utils.py#L15) with `RequestContext`.
[...]

PR #378, opened 2023-04-19
Bulk Site Creator V2, remove dependency on icommons-common and DCCSW

Chris-Thornton-Harvard:
Merging some changes from develop into the feature branch. Removes dependency on DCCSW and icommons-common
- Adds template tags from icommons-common
- Removes icommons-common and DCCSW from requirements and settings
- Update permissions to use lti_school_permissions #363
- Use modules broken out from icommons-commons #363
- Migration cleanup to remove references to old permissions module
- Adds async_operations module #355
- Adds the CanvasSchoolTemplate model to CAAT TLT-4439

Note: since the canvas_school_template table already exists, a fake migrate should be run. This will insert a record into the django_migrations table for the model but not apply any changes
python manage.py migrate  canvas_account_admin_tools --fake


canvas_account_admin_tools/templatetags/django_helpers.py
```diff
@@ -0,0 +1,17 @@
...
+
+register = template.Library()
+
+
+@register.simple_tag
+def settings_value(name):
+    return getattr(settings, name, "")
+
+
+@register.filter(is_safe=True)
```
cmurtaugh, line 15: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It looks like you have set the `is_safe` flag to `True` for a custom filter. The `is_safe` flag tells Django not to apply escaping on the value returned by the filter even though the input is escaped. This could expose your application to cross-site scripting (XSS) vulnerabilities. We recommend that you remove `is_safe` and use `mark_safe()` to explicitly mark safe content.

[Learn more](https://cwe.mitre.org/data/definitions/79.html)

cmurtaugh (commented):
Looks good -- the only part I have a question about is why the CanvasSchoolTemplate model is placed under `common` rather than in the tool app.

Chris-Thornton-Harvard:
I originally placed it in common as that dir was housing things that the canvas_site_creator and bulk_site_creator apps shared but it probably is better placed at the project level. 
I moved the model into canvas_account_admin_tools/models.py and updated import references

Chris-Thornton-Harvard:
Note: since the canvas_school_template table already exists, a fake migrate should be run. This will insert a record into the django_migrations table for the model but not apply any changes
`python manage.py migrate  canvas_account_admin_tools --fake `

PR #379, opened 2023-04-20
TLT-4443 Set and filter bulk processing flag

Chris-Thornton-Harvard:
Adds bulk_processing field to CourseInstance queryset filters to filter out potential course instances that are currently being processed by a bulk job.

Set bulk_processing to true for each course instance prior to submitting a bulk job.

To test, create a new bulk job via the tool. Upon successful submission of job, go to new job page with the same filters that were used for the bulk job (term + dept/cg) and there should be no results returned.

### DIFF
--- a/bulk_site_creator/views.py
+++ b/bulk_site_creator/views.py
@@ -132,10 +132,9 @@ def new_job(request):
 
         # Retrieve all course instances for the given term_id and account that do not have Canvas course sites
         # nor are set to be fed into Canvas via the automated feed
-        # TODO Add bulk_processing flag to filter
         potential_course_sites_query = get_course_instance_query_set(
             selected_term_id, sis_account_id
-        ).filter(canvas_course_id__isnull=True, sync_to_canvas=0)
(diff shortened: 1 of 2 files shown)


bulk_site_creator/views.py
```diff
@@ -162,7 +161,6 @@ def new_job(request):
 @lti_permission_required(settings.PERMISSION_BULK_SITE_CREATOR)
 @require_http_methods(["POST"])
 def create_bulk_job(request: HttpRequest) -> Optional[JsonResponse]:
```
cmurtaugh, line 163: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This function calls 20 other functions. By comparison, 99% of the functions in the CodeGuru reference dataset use fewer. This indicates the function is highly coupled with other functions. A function that is highly coupled with other functions is difficult to understand and its behavior might change unexpectedly when one of its referenced functions is updated. High coupling could also increase the integration test complexity, maintenance cost and technical debt. We recommend that you simplify this function or break it into multiple functions.

PR #380, opened 2023-04-20
Task/haydn9000/tlt 4442/remove column add refresh btn

haydn9000:
- Removed Job ID column and replaced with a clickable eye icon to be brought to job detail page.
- Add refresh button on index table
- Add refresh button on job detail table

Note:
Added job ID data attribute to clickable (view job detail) eye icon for convenience, in case dev needs to see job id when inspecting the eye icon element. Can remove before merge if not needed.
It looks like this when inspecting element:
job id data attribute screenshot

View job detail clickable eye icon:
clickable eye icon image


Chris-Thornton-Harvard:
👍

PR #381, opened 2023-04-24
Task/haydn9000/tlt 4449/fix coursegroup and department not filtering

haydn9000:
- Fixed Course Group and Department filter not filtering potential course sites in new_job template/view 
- Formatted timestamps displayed in index table.
- Fixed Started by column not populated in index table.

Note:

- Tested (locally) filtering potential course sites in new_job template/view in account -> Harvard University > Harvard College/GSAS
  - Selecting term 2023 Spring / January, and Spanish for Course Group should return test data.
  
- And tested in account -> Harvard University > Harvard Divinity School
  - Selecting term Spring 2023, and ECON for Department should return test data.

- Deployed in DEV

### DIFF
--- a/bulk_site_creator/templates/bulk_site_creator/new_job.html
+++ b/bulk_site_creator/templates/bulk_site_creator/new_job.html
@@ -38,12 +38,12 @@
                             <br>
                             <select id="courseDepartment" name="courseDepartment" class="form-control">
                                 {% if not selected_department_id or selected_department_id == '0' %}
-                                    <option value="0" selected>Department</option>
+                                    <option value="dept:0" selected>Department</option>
                                 {% else %}
-                                    <option value="0">Department</option>
+                                    <option value="dept:0">Department</option>
                                 {% endif %}
                                 {% for department in departments %}
-                                    {% if selected_department_id and selected_department_id == department.id %}
+                                    {% if selected_department_id and "dept:"|add:selected_department_id == department.id %}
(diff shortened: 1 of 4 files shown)


PatrickMTonne (approved):
👍

cmurtaugh (commented):
I'm not sure I understand why we're adding the `dept:` or `coursegroup:` prefix and then just splitting it off later.

bulk_site_creator/templates/bulk_site_creator/new_job.html
```diff
@@ -37,13 +37,13 @@ <h3 class="breadcrumbs">
...
-                                    <option value="0" selected>Department</option>
+                                {% if not selected_department_id or selected_department_id == '0' %}
+                                    <option value="department:0" selected>Department</option>
                                 {% else %}
-                                    <option value="0">Department</option>
+                                    <option value="department:0">Department</option>
                                 {% endif %}
                                 {% for department in departments %}
-                                    {% if selected_department_id and selected_department_id == department.id %}
+                                    {% if selected_department_id and "dept:"|add:selected_department_id == department.id %}
```
cmurtaugh, line 46: I realize that we're throwing away everything before the `:` but it's a little confusing to use `dept` here but `department` for the option above.

haydn9000: Ah that's a good catch, those are typos. I will update

PR #383, opened 2023-05-04
Only allow access to the Bulk Enrollment tool at the school-account level

cmurtaugh:
This PR limits access to the Bulk Enrollment tool to school-level sub-accounts. 

The dashboard card will be hidden if the launching sub-account SIS ID doesn't start with `school:`. If the user were to somehow request the Bulk Enrollment index page, they will get an error message. 

This PR adds a generic `error.html` template that can be rendered by any view in this project. The view context dict can contain `error_title` and `error_message` keys which will be used to render the error page.


Chris-Thornton-Harvard:
👍

PR #389, opened 2023-05-12
Task/haydn9000/tlt 4458/update UI ux select btn

haydn9000:
Note: deployed in QA. [Test account here](https://harvard.test.instructure.com/accounts/39).

- One button with text changing if items are selected in the table.
    -  adfafd

- Show no courses for selected term and department/CG, when there is no result.
    - sadafsd

- Disable button and display spinner in confirmation modal after clicking yes.
    - jhggjw54

### DIFF
--- a/bulk_site_creator/templates/bulk_site_creator/new_job.html
+++ b/bulk_site_creator/templates/bulk_site_creator/new_job.html
@@ -157,7 +157,7 @@
               </div>
             </div>
             <!-- If no potential_course_sites -->
-            {% else %}
+            {% elif selected_term_id %}
                 <hr>
                 <div class="row">
                     <h4>No courses ready for Canvas course site creation</h4>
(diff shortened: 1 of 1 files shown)


bulk_site_creator/templates/bulk_site_creator/new_job.html
```diff
@@ -157,6 +156,12 @@ <h5 class="modal-title" id="staticBackdropLabel">Confirm Canvas Course Site Crea
                 </div>
               </div>
             </div>
+            <!-- If no potential_course_sites -->
+            {% else %}
+                <hr>
+                <div class="row">
+                    <h4>No courses ready for Canvas course site creation</h4>
```
Chris-Thornton-Harvard, line 163: Maybe conditionally display this block so it doesn't display on the initial load of the page. You could do a check something along the lines of {% if selected_term_id %}, because that context value will only be provided after at least one search has been conducted.
Otherwise it could appear that there are no courses available for the term + dept/coursegroup that is displayed on the initial load of the page

haydn9000: Good idea, I'll do that!

Chris-Thornton-Harvard:
Looks great! 👍

PR #395, opened 2023-05-23
Only try to extract params from the request body if body is not empty

cmurtaugh:
This section of the code tries to extract parameters from API requests that are proxied through CAAT to the iCommons API in order to log them. 

Some part of the frontend JS code makes a lot of PATCH or POST requests with an empty body which was breaking the `json.loads()` call at line 223. This hotfix just checks to make sure that the body is not empty before we try to load it.


haydn9000 (approved):
👍

sapnamysore:
👍

PR #396, opened 2023-05-31
TLT-4434 Create test courses management command

Chris-Thornton-Harvard:
Adds a new management command to the bulk_site_creator app in CAAT

This command can be used to 
- Create new Courses and Course Instances for a given school
- Create new course instances for a given Course

If course_id or term_id arguments are not provided, then a school argument must be provided to the command.

The default amount of Course Instances created with each execution is 15 but can be overridden with the --amount argument.

### DIFF
--- a/bulk_site_creator/management/commands/create_test_courses.py
+++ b/bulk_site_creator/management/commands/create_test_courses.py
@@ -1,0 +1,61 @@
+import logging
+import random
+import string
+
+from coursemanager.models import Course, CourseInstance, Term
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Management command to help populate the COURSEMANAGER DB with test Courses and Course Instances.
+    If course_id or term_id arguments are not provided, then a school argument must be provided to the command.
+    """
+    help = 'Creates course_instance records in the coursemanager DB to be used in bulk site creation testing'
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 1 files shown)


bulk_site_creator/management/commands/create_test_courses.py
```diff
@@ -0,0 +1,61 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('--term_id', help='Term to associate new Course Instances to')
+        parser.add_argument('--course_id', help='Existing Course to associate new Course Instances to')
+        parser.add_argument('--department_id', help='Department ID to associate a new Course to')
+        parser.add_argument('--cg_id', help='Course Group ID to associate a new Course to')
+        parser.add_argument('--school', help='Lowercase abbreviation of school to associate a new Course to')
+        parser.add_argument('--amount', default=15, type=int, help='Amount of course instance records to create')
+
+    def handle(self, *args, **options):
+        if options.get('course-id', None):
```
cmurtaugh, line 27: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It appears you are using the `get()` method to retrieve the value for a key from a dictionary. If the key is not present, then `get()` returns the value that is passed in its optional second parameter. If if the key is not present, then `None` is the default return value. Therefore, there is no need to explicitly specify `None` in the second parameter.

[Learn more](https://docs.python.org/3/library/stdtypes.html#dict.get)

Similar issue at line numbers 34, 35, and 47.

haydn9000:
👍

PR #398, opened 2023-06-05
TLT-4468 - Bugfix - Jobs with no template fail

Chris-Thornton-Harvard:
Check the submitted template ID and if it is the "No template" option (value of 0), then set. the ID to None and template name to 'No template'


cmurtaugh:
We identified new issues on unchanged lines of code. Navigate to the Amazon CodeGuru Reviewer console to view the recommendations to fix them.

haydn9000 (approved):
👍

PR #401, opened 2023-06-28
TLT-4384 Single site creator app rewrite

Chris-Thornton-Harvard:
Removes Angular from the ILE/SB site creator 
Uses latest versions of front end libraries (JQuery, BootStrap)


PatrickMTonne (approved):
👍

PR #404, opened 2023-07-11
Handle IntegrityError for existing course code

haydn9000:
- Handle IntegrityError for when user tries to create a course with an existing course code.
- Formatted imports using isort

- Test GIF:
    - gif1


cmurtaugh (approved):
Looks good!

PR #405, opened 2023-07-18
Task/haydn9000/tlt 4406/remove lti permissions django app

haydn9000:
- Removed `lti_permissions` Django app
- Removed no longer needed migration files

Note:
- Deployed in Dev
- After deploying we may want to remove the database migration records for the deleted migration files to prevent future confusion.


Chris-Thornton-Harvard:
👍

PR #406, opened 2023-07-20
TLT-4481 Paginated Dynamo query

Chris-Thornton-Harvard:
If the LastEvaluatedKey is present in a DynamoDB query response, it means that there is additional information to be retrieved.

Make additional queries using that key as the value for 'ExclusiveStartKey' and add results to the tasks list

### DIFF
--- a/bulk_site_creator/views.py
+++ b/bulk_site_creator/views.py
@@ -100,7 +100,15 @@ def job_detail(request: HttpRequest, job_id: str) -> HttpResponse:
         'KeyConditionExpression': Key('pk').eq(job_id),
         'ScanIndexForward': False,
     }
-    tasks = table.query(**tasks_query_params)['Items']
+    task_query_result = table.query(**tasks_query_params)
+    tasks = task_query_result['Items']
+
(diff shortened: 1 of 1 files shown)


bulk_site_creator/views.py
```diff
@@ -100,7 +100,15 @@ def job_detail(request: HttpRequest, job_id: str) -> HttpResponse:
         'KeyConditionExpression': Key('pk').eq(job_id),
         'ScanIndexForward': False,
     }
-    tasks = table.query(**tasks_query_params)['Items']
+    task_query_result = table.query(**tasks_query_params)
+    tasks = task_query_result['Items']
+
+    # If there are additional items to be retrieved for this job, the LastEvaluatedKey will be present
+    # Us this key as the starting point for subsequent queries to build a full list
```
PatrickMTonne, line 107: `us` should be `use` I believe?

PatrickMTonne (approved):
👍 Looks like they're may be a minor typo on line 107 but looks good otherwise!

PR #407, opened 2023-07-20
TLT-4478: allow single site creation in any sub-account

PatrickMTonne:
This PR:
- Adds a dropdown to the tool's interface that allows users to select a department/coursegroup for the ILE/SB course to live in.
- Includes minor CSS updates (use `class="form-control"` instead of `class="form-select"` for all dropdowns to better indicate dropdown functionality).

Deployed and tested in QA.


haydn9000 (approved):
Looks good. Nice updating the drowpdowns CSS class to `class="form-select"` 👍

PR #408, opened 2023-08-04
Update `account_id` when creating course in COLGSAS sub-account

PatrickMTonne:
Update that came out of meeting with client services (see [Google doc](https://docs.google.com/document/d/1-qu6r8VD7GEwnC7WdGU7fIwgRQ3KCXIYnET4C2qh8zk/edit)).

This update resolves an issue where calls to create a course in the COLGSAS sub-account failed because the API endpoint was configured to search against sub-account path `dept:{account_id}` instead of (the correct) `coursegroup:{account_id}`.

### DIFF
--- a/canvas_site_creator/utils.py
+++ b/canvas_site_creator/utils.py
@@ -47,9 +47,12 @@ def create_canvas_course_and_section(data):
     template_id = data['template_id']
 
     # The course may be associated with either a department or coursegroup (depending on the school)
-    department_or_course_group_id = course.department_id if course.department_id else course.course_group_id
+    if course.department_id:
+        account_endpoint = f'dept:{course.department_id}'
+    else:
+        account_endpoint = f'coursegroup:{course.course_group_id}'
     # If this is a blueprint course, create course at school level not in the ILE sub account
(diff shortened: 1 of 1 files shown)


canvas_site_creator/utils.py
```diff
@@ -47,9 +47,12 @@ def create_canvas_course_and_section(data):
     template_id = data['template_id']
 
     # The course may be associated with either a department or coursegroup (depending on the school)
-    department_or_course_group_id = course.department_id if course.department_id else course.course_group_id
+    if course.department_id:
```
cmurtaugh, line 50: I think we might want to reverse this, and explicitly limit to `colgsas`. E.g. 
```
if course.school_id == 'colgsas' and course.course_group_id:
    # use the course group id
else:
    # use the department id
```
In the past we've erroneously received course group IDs in the DCE feed, and many/all FAS courses will have both a department and a course_group. For FAS courses we want to prefer the course_group, and for all other schools we want to ignore course_groups altogether.

PatrickMTonne: Updated in [d1d5926](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/pull/408/commits/d1d592648b1957087e7507a4346f9ff805de8487).

PR #410, opened 2023-08-09
Ensure ILE/SB sub-accounts appear in single site creator

PatrickMTonne:
This PR makes a few updates to ensure that ILE/SB sub-accounts appear in the single site creator dropdown. Conversely, ILE/SB sub-accounts have been removed from the bulk site creator dropdown.

This is accomplished by adding in an optional flag, `include_ile_sb`, to the `get_course_group_data_for_school` and `get_department_data_for_school` functions. This defaults to `True`, but can optionally be made `False` to prevent ILE/SB sub-accounts from returning in the results list.

Deployed to QA.


Chris-Thornton-Harvard:
👍

PR #411, opened 2023-08-15
Use the active_and_deleted model manager in Site Deletion tool

Chris-Thornton-Harvard:
Will get all course instance records vs the filtering of deleted=1 defined in the default objects get_queryset

When a course instance has the deleted flag as true, currently the tool will not be able to query to find and delete the entered course instance ID.


cmurtaugh:
We identified new issues on unchanged lines of code. Navigate to the Amazon CodeGuru Reviewer console to view the recommendations to fix them.

cmurtaugh (approved):

PR #412, opened 2023-08-29
Task/haydn9000/tlt 4491/fix logging issue

haydn9000:
To use our manually configured logging and receive stack trace logs in Splunk when deployed in prod, [LOGGING_CONFIG](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-LOGGING_CONFIG) has been set to None. 

Also added the `django` logger to log more messages from just the [django](https://docs.djangoproject.com/en/4.2/ref/logging/#django-logger) named logger

Note:
- _Setting [LOGGING_CONFIG](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-LOGGING_CONFIG) to None only means that the automatic configuration process is disabled, not logging itself. If you disable the configuration process, Django will still make logging calls, falling back to whatever default logging behavior is defined._ - [Django](https://docs.djangoproject.com/en/4.2/topics/logging/#disabling-logging-configuration)

- Deployed in DEV
- Tested by setting `DEBUG = False` and updated line `52` in `bulk_enrollment_tool > views.py` from `KeyConditionExpression=Key('pk').eq(f'SCHOOL#{tool_launch_school.upper()}')` to `KeyConditionExpression=Key('pk').eq(f''),`. This would give the 500 error page but send detailed logs or traceback to Splunk.

### DIFF
--- a/bulk_enrollment_tool/views.py
+++ b/bulk_enrollment_tool/views.py
@@ -49,7 +49,7 @@ def index(request):
     table = dynamodb.Table(
         settings.BULK_ENROLLMENT_TOOL_SETTINGS['bulk_enrollment_dynamodb_table'])
     response = table.query(
-        KeyConditionExpression=Key('pk').eq(""),
+        KeyConditionExpression=Key('pk').eq(f'SCHOOL#{tool_launch_school.upper()}'),
         ScanIndexForward=False,  # Records in reverse order (DESC).
         Limit=10,
     )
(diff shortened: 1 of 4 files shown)


Chris-Thornton-Harvard:
👍

cmurtaugh:
Changes were incorporated into hotfix v3.8.3.2 via PR #420.  Closing this one.

PR #413, opened 2023-08-31
TLT-4492 / TLT-4513 Use latest icommons-ui base template and latest JQuery & Bootstrap

Chris-Thornton-Harvard:


haydn9000 (approved):
👍 

I didn't know what `{{ block.super }}` was, I looked it up and that's a pretty useful variable to include stuff from parent. I'll keep that in mind

PR #416, opened 2023-09-06
Tlt 4503/idkjay/update bulk enrollment tool description

idkjay:
- Updated success message upon CSV file upload to inform user that there is a wait time before changes are reflected on Canvas
- Added the same text to the help page


Chris-Thornton-Harvard:
Looks good, my only personal input would be changing fed -> reflected, eg: `It may take up to 5 minutes for changes to be reflected in Canvas`

PR #421, opened 2023-10-19
Task/idkjay/tlt 4541/bulk course settings models and views update

idkjay:
- Job model and views changes to incorporate `meta_term_id`
- Migration made and applied on DEV for the Job model update
- Bulk Job reversion also updated with `meta_term_id`
- To change pg database schema for use in bulk_course_settings_processing lambda function


Chris-Thornton-Harvard:
This will also require a migration to be applied. 
To generate the migration file do something along the lines of `python manage.py makemigrations bulk_course_settings`
Take a look at the docs: https://docs.djangoproject.com/en/4.2/topics/migrations/

bulk_course_settings/migrations/0005_auto_20231019_1548.py
Chris-Thornton-Harvard, line 1: The `AddField` operation is correct but can you look a bit more into what is changing regarding the `AlterField` operations in this migration?

idkjay: All the `AlterField` operations seem to be reflecting changes made in `models.py` 5 years ago. I'm not totally sure why they would get added into this particular migration run..

idkjay: Colin mentioned that the `AlterField` operations were all added due to changes in Django itself between now and when the fields of those models were originally created.

Chris-Thornton-Harvard: Could you apply the migration to one env and then try running makemigrations again to see if another migration gets created? If so, we may want to order the choice fields

idkjay: No new migrations for `bulk_course_settings` was created. Here is the terminal after `python manage.py makemigrations`:
image

And here is the result of a `showmigrations` command:
image
The `0005_auto_20231019_1548` migration file was the latest one before and after the commands.

Chris-Thornton-Harvard:
👍 Looks good!

PR #422, opened 2023-10-25
Task/idkjay/tlt 4539/bulk site creator add select all checkbox

idkjay:
- Added select all checkbox 
- Added Javascript functionality

Currently, this is draft PR as this is a hard coded implementation and not using DataTables convention. The Select All checkbox is above the `Show X entries` when it should be below. 

[JIRA](https://at-harvard.atlassian.net/browse/TLT-4539)

### DIFF
--- a/bulk_site_creator/templates/bulk_site_creator/new_job.html
+++ b/bulk_site_creator/templates/bulk_site_creator/new_job.html
@@ -112,7 +112,7 @@
                 <table id="course-table" class="datatable table table-striped">
                     <thead>
                         <tr>
-                            <th><input name="selectAll" id="select-all-courses" value="all" type="checkbox" /></th>
+                            <th style="padding-left: 54px;"><input class="col-select" name="selectAll" id="select-all-courses" value="all" type="checkbox" /></th>
                             <th>SIS ID</th>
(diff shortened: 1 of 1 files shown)


Chris-Thornton-Harvard:
Tested on QA

I was able to select the checkbox which highlighted all rows in the DataTable but only checks the checkboxes on the first page.
Submitting the job did include all courses in the table.
When the select all button is toggled and then one of the items in the table deselected, the select all button remains checked. This remains checked even if all items in the table are unchecked.

Behavior should more closely resemble that in the bulk publish tool. Specifically I would take a look at the IndexController and the toggleSelectAll function, https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/793b8214886b654c935802fc3277c18b7c77247f/publish_courses/static/publish_courses/js/controllers/IndexController.js#L61

idkjay:
Built, deployed, and tested on DEV and QA

Chris-Thornton-Harvard:
Changes look good!
Only thing I noticed is the "Create All" button no longer changes to "Create Selected" when there are rows selected.
Other than that, just a nitpick, there are a couple of console log statements from debugging that maybe could be removed

PR #425, opened 2023-11-28
Task/tlt 4555/idkjay/find courses renders default

idkjay:
- Refactored logic to allow for the selected department or course group to show up as the visible select option after the Find Courses button is clicked
- Added logic to `views.py` to help facilitate the rendering of the `departments` or `course_groups` list and to retrieve the name of the selected department or course group from the Coursemanager database.

tlt4555

Note:
- Deployed on `DEV`

### DIFF
--- a/bulk_site_creator/templates/bulk_site_creator/new_job.html
+++ b/bulk_site_creator/templates/bulk_site_creator/new_job.html
@@ -35,15 +35,17 @@
                                 </label>
                                 <br>
                                 <select id="courseDepartment" name="courseDepartment" class="form-select">
-                                    {% if not selected_department_id or selected_department_id == '0' %}
-                                        <option value="0" selected>All Departments</option>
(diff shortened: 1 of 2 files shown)


Chris-Thornton-Harvard:
The reason why the selected department/course group don't match the original versions conditional check is that there is a type mismatch. On POST, the selected ID is a string and then when it comes time to render and go through the conditional checks again, it is comparing a string against an int and then skipped when it should have been a hit.

Ideally you want to minimize the amount of queries you do for efficiency and in this case it would be doing a redundant query just to get the name.

One approach would be to cast the selected values as ints if they exist, something along the lines of:
```        
if selected_course_group_id and selected_course_group_id != '0':
        selected_course_group_id = int(selected_course_group_id)
        potential_course_sites_query = potential_course_sites_query.filter(course__course_group=selected_course_group_id)```

PR #427, opened 2024-01-11
task/tlt-4569/idkjay/modify_job_details_count

idkjay:
A few fields within the Job model are not needed since we are trending towards asynchronicity. But to cover for older instances that have values for those parameters, `null=True` will be added to its required parameters. 

Fields changed:
```
details_total_count
details_success_count
details_failed_count
details_skipped_count
```

The migration was made and applied on `DEV` for the Job model.


idkjay:
image

idkjay:
I changed it from `details_total_count = models.IntegerField(default=0)` to `details_total_count = models.IntegerField(null=True)` but would `details_total_count = models.IntegerField(default=0, null=True)` be more appropriate?

Chris-Thornton-Harvard:
I think just making it nullable should be fine as having it be 0 wouldn't accurately reflect the state of the job details.
Changes look good but I would change the PR base from develop to your feature branch prior to merging

PR #428, opened 2024-01-17
TLT-4564: Update error message when deleting secondarily-crosslisted course

PatrickMTonne:
Using the Site Deletion tool, attempting to delete a secondarily-crosslisted course would provide the following message, creating confusion for users:

`Course instance ID ({course_search_term}) does not match Canvas course SIS ID ({cc["sis_course_id"]}) for Canvas course {ci.canvas_course_id}. Aborting.`

This message has been updated to better highlight why the deletion cannot proceed and steps needed to resolve the issue:

`Cannot delete course {course_search_term} as it is secondarily-crosslisted with another course. To proceed, undo the existing crosslisting and retry the deletion.`

### Testing

- Identified a secondarily-crosslisted course via the crosslisting tool
- Attempted to delete the course using the Site Deletion tool and was provided the new message

### Notes

- The primary course within a crosslisting _is_ able to be deleted. This functionality is unchanged from before.


Chris-Thornton-Harvard:
👍

PR #432, opened 2024-02-02
TLT-4568 Bulk Course Settings tweaks

Chris-Thornton-Harvard:
Code adjustments based off of changes from Jordan's branch

- Removes detail counts from Job model and associated migration. (name would conflict with annotations)
- Adds annotations to Job queryset for detail counts
- Bumps Django version


haydn9000:
Went over PR with Chris and all  looks good 👍. 
I'd say add a story for adding logic to limit queueing lambda payload to 2500 courses so we don't run into payload limits.

PR #438, opened 2024-03-07
task/tlt-4604/idkjay/limit_deletion_to_sites_under_current_sub_account

idkjay:
- Added a check to make sure you are in the right sub-account in order to delete a Canvas course
- Added a function to pull current school account id of where tool was launched
- Currently, you can delete a sub-sub-account (department or coursegroup) course whether you're in that same sub-sub-account or the sub-account above it. 
- You can also delete courses of the sub-account while inside the sub-sub-account. 
- For above two points, we can add extra protections later.

Note
- Tested and deployed on `QA`

### DIFF
--- a/canvas_site_deletion/views.py
+++ b/canvas_site_deletion/views.py
@@ -4,6 +4,9 @@ import time
 from canvas_sdk import RequestContext
 from canvas_sdk.exceptions import CanvasAPIError
 from canvas_sdk.methods import courses, sections
+from canvas_sdk.methods.courses import \
+    get_single_course_courses as canvas_get_course
+from canvas_sdk.methods.accounts import (get_sub_accounts_of_account,get_single_account)
 from canvas_sdk.utils import get_all_list_data
 from coursemanager.models import CourseInstance, CourseSite, SiteMap, Course
 from django.conf import settings
@@ -89,20 +92,19 @@ def lookup(request):
 @require_http_methods(['GET', 'POST'])
 def delete(request, pk):
     canvas_course_id = None
-    custom_canvas_account_sis_id = request.LTI['custom_canvas_account_sis_id']
     try:
         ci = CourseInstance.active_and_deleted.get(course_instance_id=pk)
 
         canvas_course_id = ci.canvas_course_id
-        canvas_site_school_id = ci.term.school.school_id
-        canvas_course_instance_id = ci.course_instance_id
         ts = int(time.time())
 
-        # if school_id of the Canvas course does not match the sub-account id, deletion will not be possible
-        tool_launch_school = get_tool_launch_school(custom_canvas_account_sis_id, canvas_course_instance_id)
+        current_tool_launch_account_id = int(request.LTI["custom_canvas_account_id"])
 
-        if canvas_site_school_id != tool_launch_school:
-            messages.error(request, f'You must be under the correct sub-account in order to delete Canvas course {canvas_course_id} and course_instance {pk}.')
+        account_ids = get_account_hierarchy(ci.canvas_course_id)
+
+        if current_tool_launch_account_id not in account_ids:
+            logger.error(f'User {request.user} is trying to delete the Canvas site associated with course_instance {pk} but is not in the correct account or sub-account where they have launched the site deletion tool.')
+            messages.error(request, f'You must be under the correct account or sub-account in order to delete Canvas course {canvas_course_id} and course_instance {pk}.')
             return render(request, 'canvas_site_deletion/index.html')
 
         if not canvas_course_id:
--- a/canvas_site_deletion/views.py
+++ b/canvas_site_deletion/views.py
@@ -103,7 +103,8 @@ def delete(request, pk):
         account_ids = get_account_hierarchy(ci.canvas_course_id)
 
         if current_tool_launch_account_id not in account_ids:
-            logger.error(f'User {request.user} is trying to delete the Canvas site associated with course_instance {pk} but is not in the correct account or sub-account where they have launched the site deletion tool.')
+            logger.error(f'User {request.user} is trying to delete the Canvas site associated with course_instance {pk} but is not in the correct account or sub-account where they have launched the site deletion tool.', 
+                         extra={"course_instance_id": pk, "account_ids": account_ids, "user_id": request.user})
             messages.error(request, f'You must be under the correct account or sub-account in order to delete Canvas course {canvas_course_id} and course_instance {pk}.')
             return render(request, 'canvas_site_deletion/index.html')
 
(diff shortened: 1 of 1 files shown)


canvas_site_deletion/views.py
```diff
@@ -161,3 +172,25 @@ def delete(request, pk):
...
+    try:
+        tool_launch_school = "unknown"  # Initialize tool_launch_school
+        if custom_canvas_account_sis_id.startswith('coursegroup:'):
+            tool_launch_school = "colgsas"
+        elif custom_canvas_account_sis_id.startswith('dept:'):
+            dept_id = custom_canvas_account_sis_id.split(':')[1]
+            ci = CourseInstance.objects.get(course_instance_id=ci_course_instance_id)
+            course_id = ci.course_id
+            course = Course.objects.get(course_id=course_id, department_id=dept_id)
+            tool_launch_school = course.school_id
```
cmurtaugh, line 190: This looks wrong to me -- since this could return the school_id associated with the course_instance, it could allow someone to accidentally delete a course that they should not be able to delete.

cmurtaugh (changes requested):
I think the logic is a bit flawed; instead of always checking whether the account associated with the course matches the account where the tool was launched, in some cases it would compare the account associated with the course with itself (and get a false positive). 

It would be better to just compare the course's location in the Canvas account hierarchy with the account where the tool was launched. I'd do that using the Canvas API rather than the coursemanager database.

canvas_site_deletion/views.py
```diff
@@ -161,3 +174,33 @@ def delete(request, pk):
...
+
+    # use course id to get account id
+    canvas_course = canvas_get_course(SDK_CONTEXT, course_id).json()
+    canvas_course_account_id = canvas_course["account_id"] 
+    account_ids.append(canvas_course_account_id)
+
+    # traverse account hierarchy in both directions
+    # first downwards to find sub-accounts
+    sub_accounts = get_sub_accounts_of_account(SDK_CONTEXT, canvas_course_account_id).json()
+    for sub_account in sub_accounts:
```
cmurtaugh, line 193: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

To create a `list`, try to use `list` comprehension instead of a loop. List comprehension is the preferred way to make a list using Python, and it&#39;s simpler and easier to understand than using a loop.

[Learn more](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions)

haydn9000 (commented):
Looks good besides the type hint stuff above. Also tested in QA and it seems to be working. 👍

My testing steps:
- I went to the HUIT AcTS. > Admin Console > Canvas Site Deletion
    - I then entenred this course instance id `571039` (HMS course).
    - Then I got the error message "You must be under the correct account..."
- I then went to Harvard Divinity School > OMS > Admin Console >  Canvas Site Deletion
    - Entered this course instance id `451037` to try to delete this course that is from `Harvard Divinity School > HDS` account.
    - Then I got the error message "You must be under the correct account..."
    - I then went to Harvard Divinity School > Admin Console > Canvas Site Deletion
    - I then entenred this course instance id `451037` (course from Harvard Divinity School > HDS account).
    - Then I was able to get the success message "Successfully deleted Canvas course..."

haydn9000 (approved):
👍

canvas_site_deletion/views.py
```diff
@@ -161,3 +174,33 @@ def delete(request, pk):
...
+    account_ids = []
+
+    # use course id to get account id
+    canvas_course = canvas_get_course(SDK_CONTEXT, course_id).json()
+    canvas_course_account_id = canvas_course["account_id"] 
+    account_ids.append(canvas_course_account_id)
+
+    # traverse account hierarchy in both directions
+    # first downwards to find sub-accounts
+    sub_accounts = get_sub_accounts_of_account(SDK_CONTEXT, canvas_course_account_id).json()
```
Chris-Thornton-Harvard, line 192: There is a `recursive` argument for this call:
`If true, the entire account tree underneath this account will be returned (though still paginated). If false, only direct sub-accounts of this account will be returned. Defaults to false.`

Chris-Thornton-Harvard:
Wouldn't you want to use the account ID of where the tool was launched (`current_tool_launch_account_id`) and then get the list of subaccount ID's from that and then check if the input Canvas course's account ID is in that list?

idkjay:
> Wouldn't you want to use the account ID of where the tool was launched (`current_tool_launch_account_id`) and then get the list of subaccount ID's from that and then check if the input Canvas course's account ID is in that list?

I wanted to create a function that, given a course ID, returns an array of the account IDs that it descends from. That way you could quickly check to see if the launch account is in that list. Would you say it would be more efficient giving the function the `current_tool_launch_account_id`?

cmurtaugh:
Logically either way could work, but starting with the course's account ID and working up the tree should be more efficient than starting with the launch account ID and working down. E.g. there are hundreds of sub-accounts underneath COLGSAS, but starting from a course and working up it'd only take one or two hops to get to COLGSAS.

canvas_site_deletion/views.py
```diff
@@ -161,3 +174,33 @@ def delete(request, pk):
...
+    Get the list of accounts associated with a given course ID
+    """
+    account_ids = []
+
+    # use course id to get account id
+    canvas_course = canvas_get_course(SDK_CONTEXT, course_id).json()
+    canvas_course_account_id = canvas_course["account_id"] 
+    account_ids.append(canvas_course_account_id)
+
+    # traverse account hierarchy in both directions
```
cmurtaugh, line 190: We don't want to traverse the hierarchy in both directions here -- we only want to look up the hierarchy. If we were to also look down the hierarchy, that would mean that a department-level admin would be able to delete a course at the school level which we don't want.

canvas_site_deletion/views.py
```diff
@@ -91,9 +94,19 @@ def delete(request, pk):
...
+
         canvas_course_id = ci.canvas_course_id
         ts = int(time.time())
 
+        current_tool_launch_account_id = int(request.LTI["custom_canvas_account_id"])
+
+        account_ids = get_account_hierarchy(ci.canvas_course_id)
+
+        if current_tool_launch_account_id not in account_ids:
+            logger.error(f'User {request.user} is trying to delete the Canvas site associated with course_instance {pk} but is not in the correct account or sub-account where they have launched the site deletion tool.')
```
cmurtaugh, line 106: Let's add the account_ids list to the log message too so that it's easier to debug. You can also add a second argument to the `logger.error()` call like this: 
```python
logger.error(f"error message...", extra={"course_instance_id": pk, "account_ids": account_ids});
```

PR #443, opened 2024-04-02
TLT-4635: Handle filtering when using Select All checkbox

PatrickMTonne:
## Summary
This PR:
- Updates the Select All checkbox to account for any filters applied by the user (via the Search input). Now, if a user applies a filter to the data set, the Select All checkbox will only select that subset of rows (as opposed to the total, unfiltered data set). This was not previously the case.

## Testing

This branch has been deployed to dev + qa.

Performed two test runs to ensure that only the selected rows are submitted within the batch job.

Test run 1 - submit filtered job (20 rows). 
- Expected 20 courses; resulted in 20 created courses

Screen Shot 2024-04-02 at 2 27 49 PM

Test run 2 - submit unfiltered job (312 rows).
- Expected 312 courses; resulted in 312 created courses

Screen Shot 2024-04-02 at 2 32 50 PM

### DIFF
--- a/bulk_site_creator/templates/bulk_site_creator/new_job.html
+++ b/bulk_site_creator/templates/bulk_site_creator/new_job.html
@@ -210,32 +210,44 @@
                 setTimeout(updateButtonText, 125); // 1/8 second wait
             });
 
-            // Select All functionality
-            $('.selectAll').on('change', function() {
+            $('.selectAll').on('change', function(event) {
+                element = this;
+            });
+
+            $('.selectAll').on('click', function(event) {
+//                event.preventDefault();
+//                var isIndeterminate = this.checked
                 if(this.checked) {
                     selectRows(table);
+                    updateSelectAllCheckbox(table);
                 } else {
                     deselectRows(table);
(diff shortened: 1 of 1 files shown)


Chris-Thornton-Harvard:
Noticed this when testing:

1. Select a few items from table with no filter. Change the ordering of any column and any selected items are now deselected.  

Both order and search events tigger a draw() and on each draw the table is re-searched. 
https://datatables.net/reference/event/order
https://datatables.net/reference/event/search
Also you may need to include dt namespace in the event listener.

As for the search field event deselecting items, this might be a good UX discussion with the team. 
To me it would not seem obvious that entering a value would deselect any items I already selected.
I also am imagining a situation where someone would want to select a few courses from a large list and it would be quickest to use the search field to put together their list for the job rather than having to manually scroll through multiple pages to find the desired courses.

haydn9000 (approved):
👍 tested the functionality in dev and it looks good!

Chris-Thornton-Harvard:
👍 Looks good! Good job with the indeterminate state!

idkjay:
👍

PR #263, opened 2019-03-12
TLT-3570 Clears the SIS ID when a site is disassociated

Chris-Thornton-Harvard:
When a course site is disassociated, we now clear the SIS ID of the Canvas course

https://jira.huit.harvard.edu/browse/TLT-3570

I used the COLGSAS course 354849 to test this.
Open that course instance and then click on associated sites tab.
Open one of the associated sites in a new tab and go to the settings page.
On the previous tab, select "Disassociate site".
You should see one less associated site.
On the other tab, refresh the page and you should see that the SIS ID field is cleared out

PR #445, opened 2024-06-10
Task/idkjay/tlt 4665/set date restrictions

idkjay:
- Added start and end dates to self enrollment course model
- Completed migration for said model update
- Added input boxes for start and end date when creating self enrollment link and wording to let use know it is optional
- Added validation logic for start and end dates
- Didn't add a dedicated error page but displayed errors when self enrollment creation did not get through date validation and redirected page back to Add New page
- Added start and end date displays to the index UI

- Deployed in `DEV`

image
image
image
image


self_enrollment_tool/templates/self_enrollment_tool/base.html
```diff
@@ -16,6 +16,12 @@
 
 {% block javascript %}
     {{ block.super }}
+    <!-- jQuery and Bootstrap JS -->
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/popper.js@1.16.1/dist/umd/popper.min.js"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/js/bootstrap.min.js"></script>
+
```
Chris-Thornton-Harvard, line 23: The [icommons_ui/base_lti.html](https://github.com/Harvard-University-iCommons/django-icommons-ui/blob/master/icommons_ui/templates/icommons_ui/base_lti.html) template already has the following:
```
<script src="https://static.tlt.harvard.edu/shared/jquery-3.6.3/jquery-3.6.3.min.js"></script>
<script src="https://static.tlt.harvard.edu/shared/bootstrap-5.3.0/js/bootstrap.bundle.min.js"></script>
```
Do you need these versions for the date popper to work?
Do we already have a version of the date popper in our static S3 bucket? If not we should add it there and reference it as part of the static.tlt.harvard.edu/shared/ endpoint

Chris-Thornton-Harvard:
When creating a new self enrollment link with a start date of today, I get the "Start date cannot be in the past.” error message

self_enrollment_tool/templates/self_enrollment_tool/base.html
idkjay: I added these to get the Jquery copy link and delete icons to work properly. The date popper was working properly before these were added.

idkjay:
@Chris-Thornton-Harvard While I wasn't able to recreate that error, would there be any situations where a user might actually need to create a new self enrollment link with a start date of today? If not, I could change the date popper to remove the option to click Today.

Chris-Thornton-Harvard:
I think we will want to break out the copy/delete issues into its own separate Jira bug report as injecting older versions of JQuery and Bootstrap are not addressing the root cause.
I spent a little time poking around the project and it looks like we are using the JQuery tooltip function vs Bootstrap which should require the `jquery-ui` package with something along the lines of 
```
<script
  src="https://code.jquery.com/ui/1.13.2/jquery-ui.min.js"
  integrity="sha256-lSjKY0/srUM9BE3dPm+c4fBo1dky2v27Gdjm2uoZaL0="
  crossorigin="anonymous"></script>
```
This got the copy function working but still no tooltip and the delete functionality is still broken.

As for Bootstrap popper & tooltips, see the following:
https://getbootstrap.com/docs/5.0/components/tooltips/
```
Tooltips rely on the 3rd party library [Popper](https://popper.js.org/) for positioning. You must include [popper.min.js](https://cdn.jsdelivr.net/npm/@popperjs/core@2.9.2/dist/umd/popper.min.js) before bootstrap.js or use bootstrap.bundle.min.js / bootstrap.bundle.js which contains Popper in order for tooltips to work!
```

self_enrollment_tool/models.py
```diff
@@ -14,5 +14,5 @@ class SelfEnrollmentCourse(models.Model):
     end_date = models.DateTimeField(null=True, blank=True)
 
     class Meta:
-        unique_together = ('course_instance_id', 'role_id',)
+        unique_together = ('course_instance_id', 'role_id', 'start_date', 'end_date')
```
Chris-Thornton-Harvard, line 17: Did we decide on removing this unique constraint altogether?

self_enrollment_tool/views.py
```diff
@@ -217,44 +217,57 @@ def enable(request, course_instance_id):
...
+                    uuid = str(uuid4())
+                    SelfEnrollmentCourse.objects.create(
+                        course_instance_id=course_instance_id,
+                        role_id=role_id,
+                        updated_by=str(request.user),
+                        uuid=uuid,
+                        start_date=start_date,
+                        end_date=end_date
+                    )
+                    logger.debug(f'Successfully saved Role_id {role_id} for Self Enrollment in course {course_instance_id}. UUID={uuid}')
```
Chris-Thornton-Harvard, line 245: Might make this an info level log since it is a CRUD operation

self_enrollment_tool/views.py
```diff
@@ -217,44 +217,57 @@ def enable(request, course_instance_id):
...
-                    messages.success(request, f'Self-Unenrollment tool already installed into the Canvas course.')
+                course_instance = SelfEnrollmentCourse.objects.filter(
+                    course_instance_id=course_instance_id,
+                    role_id=role_id,
+                    start_date=start_date,
+                    end_date=end_date
+                ).first()
+
+                if course_instance:
+                    logger.error(f'Self Enrollment is already enabled for this course {course_instance_id} with the same role and date range.')
```
Chris-Thornton-Harvard, line 232: This might be better at an info level

self_enrollment_tool/views.py
```diff
@@ -217,44 +217,57 @@ def enable(request, course_instance_id):
...
-                result = install_unenrollment_tool(course_instance_id=course_instance_id, client_id=self_unenroll_client_id)
-                if result == 'installed':
-                    messages.success(request, f'Successfully installed self-unenrollment tool into the Canvas course.')
-                elif result == 'already_installed':
-                    messages.success(request, f'Self-Unenrollment tool already installed into the Canvas course.')
+                course_instance = SelfEnrollmentCourse.objects.filter(
+                    course_instance_id=course_instance_id,
+                    role_id=role_id,
+                    start_date=start_date,
+                    end_date=end_date
```
Chris-Thornton-Harvard, line 228: Just want to clarify which direction we chose and if this should be an is_null check or not

idkjay: we will be going in the direction of is_null check

self_enrollment_tool/views.py
```diff
@@ -186,23 +190,58 @@ def enable(request, course_instance_id):
...
+                course_instance = SelfEnrollmentCourse.objects.get(
+                    course_instance_id=course_instance_id, 
+                    role_id=role_id, 
+                    start_date__isnull=True,
+                    end_date__isnull=True
+                )
+                logger.info(f'Self Enrollment is already enabled for this course  {course_instance_id} ')
                 messages.error(request, f'Self Enrollment is already enabled for this course (SIS ID {course_instance_id}) and role ({role_name}). '
                               f'See below for the previously-generated link.')
                 path = reverse('self_enrollment_tool:enroll', args=[course_instance.uuid])
```
Chris-Thornton-Harvard, line 233: This seems to throw an error which is caught and displayed to the UI as a message in the #263 exception block
`Error creating self enrollment record for course 696751 with role 10. Error details=Reverse for 'enroll' not found. 'enroll' is not a valid view function or pattern name.`

Chris-Thornton-Harvard: Screen Shot 2024-07-10 at 9 35 44 AM

Chris-Thornton-Harvard: Looking at [URLs](https://github.com/Harvard-University-iCommons/canvas_account_admin_tools/blob/develop/self_enrollment_tool/urls.py) you might want to switch that out for `enable` and possibly put it in the message since the `path` var is not being used anywhere else

Chris-Thornton-Harvard:
👍

PR #447, opened 2024-07-01
Bulk site deletion management command

Chris-Thornton-Harvard:
Adds a new management command in the Canvas site deletion app that takes a CSV containing course instance IDs and deletes the Canvas site & sections as well as clean up the CourseInstance records in our DB.


haydn9000 (approved):
Looks good! 👍

PR #448, opened 2024-08-22
Fix copy/delete buttons and tooltips

idkjay:
- Updated bootstrap attributes on elements
- Updated logic behind modals, tooltips, and buttons
- Deployed to dev


cmurtaugh:
The copy button works great! The delete button isn't working for me though. There seems to be something funny going on with the tooltip on the delete button -- it flickers if I move my pointer over the icon -- not sure if this is related to the button not functioning.

Chris-Thornton-Harvard:
Redeployed and tested latest, looks good!

PR #450, opened 2024-09-16
Cross Listing Tool switch button is broken

idkjay:
- Fixed switch button on Cross Listing Tool by putting the `<i>` inside a `<a>` and updating the font-awsome class
- Deployed on DEV
 
image


Chris-Thornton-Harvard:
👍

PR #451, opened 2024-09-16
TLT-4743: Reduce `maxlength` of `course-code` input to 46

PatrickMTonne:
Resolves [TLT-4743](https://at-harvard.atlassian.net/browse/TLT-4743).

Tested in dev/qa. Reducing the max allowed input size to 46 will ensure that, after backend processing, the value inserted into the DB column is <= 50 characters.


idkjay:
👍

Chris-Thornton-Harvard:
👍

PR #452, opened 2025-01-09
TLT-4899: Revert job status stuck in queued state

PatrickMTonne:
## Summary

Resolves [TLT-4899](https://at-harvard.atlassian.net/browse/TLT-4899).

This PR:
- Ensures that when reverting a bulk job, we send over the ID of the new job to the Lambda (as opposed to the ID of the original job that we're reverting). That way, the state of the reversion can be accurately tracked.
- Disables the Revert Job button when all courses in a job were skipped. This is to prevent an issue where a reversion of this kind of job gets stuck in a "Queued" state.

Screenshot 2025-01-09 at 6 23 56 PM

## Testing
This branch was deployed/tested on QA.

Testing was performed via creating a new bulk settings job, then reverting that job.

### DIFF
--- a/bulk_course_settings/templates/bulk_course_settings/job_detail.html
+++ b/bulk_course_settings/templates/bulk_course_settings/job_detail.html
@@ -54,7 +54,7 @@
                     {% endif %}
                 </div>
                 <div class="col-sm-2">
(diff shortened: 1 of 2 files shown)


haydn9000 (approved):
Looks good 👍

Chris-Thornton-Harvard (approved):
👍

